PR #4, opened 2020-03-06
Initial reconcile implementation

machacekondra:


pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,168 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+			_, err := r.kubeClient.CdiClient().CdiV1alpha1().DataVolumes("default").Create(&dv) // FIXME:
 			if err != nil {
 				return reconcile.Result{}, err
 			}
+		}
 
-			// Pod created successfully - don't requeue
-			return reconcile.Result{}, nil
-		} else if err != nil {
+		// Create kubevirt VM from oVirt VM:
```
pkliczewski, line 159: Would be great to move vm creation to be first. UI assumes it will be created very fast.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,168 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
-					Source: cdiv1.DataVolumeSource{
-						/*
-							Imageio: &cdiv1.DataVolumeSourceImageIO{
-								URL:           secret.Data["apiUrl"],
-								DiskID:        vm.MustDiskAttachments().Slice()[0].MustDisk().MustId(),
-								SecretRef:     "",
-								CertConfigMap: "",
-							},
-						*/
+				PVC: &corev1.PersistentVolumeClaimSpec{
```
pkliczewski, line 249: TODO: Would be great to add `storageClassName` which we should get from the mapping.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
```diff
@@ -67,7 +67,7 @@ type VirtualMachineImportOvirtSourceVMSpec struct {
 // +k8s:openapi-gen=true
 // +optional
 type VirtualMachineImportOvirtSourceVMClusterSpec struct {
-	ID string `json:"id"`
+	ID string `json:"id,omitempty"`
```
masayag: pls add comment above ID, also change the type to *string

`// +optional`

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -57,7 +65,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 	// TODO(user): Modify this to be the types you create that are owned by the primary resource
 	// Watch for changes to secondary resource Pods and requeue the owner VirtualMachineImport
-	err = c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{
+	err = c.Watch(&source.Kind{Type: &kubevirtv1.VirtualMachine{}}, &handler.EnqueueRequestForOwner{
```
masayag, line 73: Don't we need to watch also the created DataVolumes by this CR ?
This might be tricky: we create the DataVolumes separately from the VM, and in order to watch them we'd need the ability to set rule for that. I wouldn't use this CR as owner-reference for the DataVolumes, since the CR can be deleted while DataVolumes remains, but maybe a label indicating `created-by: this-cr`.

We need that so we can know when to run the VM (if needed) after the disks are created.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
-			err := r.client.Create(context.TODO(), dv)
-			if err != nil {
-				return reconcile.Result{}, err
-			}
-		}
-	*/
-
-	// Create kubevirt VM from oVirt VM:
+	diskAttachmentsLink, _ := vm.DiskAttachments()
+	diskAttachments, _ := ovirtConnection.FollowLink(diskAttachmentsLink)
```
masayag, line 136: diskAttachmentsLink and diskAttachments can be pushed into `newDVForCR` since they aren't being a concern of the reconcile loop and can be encapsulated. The signature will get  `newDVForCR(vm, ovirtSecret)`.
This might be extended to receive also the mapping of ovitr2kubevirt resources config-map.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+	dvs := newDVForCR(diskAttachments.(*ovirtsdk.DiskAttachmentSlice), ovirtSecret)
 	vmSpec := newVMForCR(vm, dvs, ovirtConnection)
-	err = r.client.Create(context.TODO(), vmSpec)
-	if err != nil {
+
+	// Set VirtualMachineImport instance as the owner and controller
+	if err := controllerutil.SetControllerReference(instance, vmSpec, r.scheme); err != nil {
```
masayag, line 141: I think we'd like to remove that ownership relation once the VM is ready, so the VM lifecycle wouldn't be tied to the VMImport CR.
VMImport CR is sort of a 'ticket' for creating the VM, but once the 'ticket' is closed, we'd like to remove this CR, while the VM should remain in our system. By setting the owner-reference of the VM for the VMImport CR, we're instructing the garbage collector to remove this VM upon cascade.

Same goes for the DVs, which make sense to mark as owned by the VM.

Thoughts ?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
-				return reconcile.Result{}, err
-			}
-		}
-	*/
-
-	// Create kubevirt VM from oVirt VM:
+	diskAttachmentsLink, _ := vm.DiskAttachments()
+	diskAttachments, _ := ovirtConnection.FollowLink(diskAttachmentsLink)
+	dvs := newDVForCR(diskAttachments.(*ovirtsdk.DiskAttachmentSlice), ovirtSecret)
 	vmSpec := newVMForCR(vm, dvs, ovirtConnection)
```
masayag: ovirtConnection isn't used in this method. is it needed ?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 		return reconcile.Result{}, err
```
masayag, line 132: I'd organize this differently, assuming not missing logic here: We should connect to ovirt only once (ideally, if connection succeeded and targetVM exists) during the invocations of the reconcile method, since we need to fetch the definitions once, verify them and use them for creating the vm and the resources (DV/PVC).

Once the VM is created, we should set in the VMIMport CR status the name of the VM as `targetVirtualMachineName` - since the name could be generated/altered from the name on ovirt.

Then we should rely on it for further queries. I assume it is legit to rely on name from status.
If we'll find a need to let the user specify the VM name, we'll have to extend the CR for that.
(Alternately, we can rely on entry in the mappings to specify the name, however this seems like an item for the VMImport.spec.targetVmName )

Once we have the VM created and the Namespace as part of the CR, there is no longer need to invoke API calls for ovirt.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
-	// Set VirtualMachineImport instance as the owner and controller
-	/*
-		if err := controllerutil.SetControllerReference(instance, pod, r.scheme); err != nil {
-			return reconcile.Result{}, err
-		}
+	// Check if this VM already exists
+	found := &kubevirtv1.VirtualMachine{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: vmSpec.Name, Namespace: vmSpec.Namespace}, found)
+	if err != nil && errors.IsNotFound(err) {
+		reqLogger.Info("Creating a new VM", "VM.Namespace", vmSpec.Namespace, "VM.Name", vmSpec.Name)
```
masayag, line 149: I think we should rely on `VMIMport.Status.targetVirtualMachineName` instead of `vmSpec.Name `since the name might be different than the VmName on ovirt due to different naming restrictions.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
 
-		// Check if this Pod already exists
-		found := &corev1.Pod{}
-		err = r.client.Get(context.TODO(), types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, found)
-		if err != nil && errors.IsNotFound(err) {
-			reqLogger.Info("Creating a new Pod", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
-			err = r.client.Create(context.TODO(), pod)
+		// Create kubevirt datavolume from oVirt VM disks:
```
masayag, line 151: Once we'll change the way we examine the existence of VM, we can push lines 130-138 to here and create the VM Spec and DVs only if the VM wasn't created and save some api calls to the engine.

If the VM is created, we should monitor the status of the DVs creation.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+		for _, dv := range dvs {
+			_, err := r.kubeClient.CdiClient().CdiV1alpha1().DataVolumes("default").Create(&dv) // FIXME:
```
masayag: The name space should be the same as the VM/VMImport CR ?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
 		return vms.Slice()[0], nil
 	}
-	return nil, fmt.Errorf("Virtual machine %v not found", vmSpec.Name)
+	return nil, fmt.Errorf("Virtual machine %v not found", *vmSpec.Name)
 }
 
 func (r *ReconcileVirtualMachineImport) fetchOvirtSecret(vmImport *v2vv1alpha1.VirtualMachineImport) (*corev1.Secret, error) {
 	secret := &corev1.Secret{}
-	err := r.client.Get(context.TODO(), types.NamespacedName{Name: vmImport.Spec.Source.Ovirt.SecretName, Namespace: vmImport.Namespace}, secret)
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: vmImport.Spec.Source.Ovirt.ProviderCredentialsSecret.Name, Namespace: vmImport.Namespace}, secret) // FIXME:
```
masayag: Secret's namespace should be taken from the following element (if provided - it's optional. if not provided, use same as VMImport CR):

```
      providerCredentialsSecret: # A secret holding the access credentials to ovirt, see example ovirt-mapping-example.yaml
        name: my-secret-with-ovirt-credentials
        namespace: default # optional, if not specified, use CR's namespace
```

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
 	return ovirtsdk.NewConnectionBuilder().
 		URL(string(secret.Data["apiUrl"])).
 		Username(string(secret.Data["username"])).
 		Password(string(secret.Data["password"])).
-		Insecure(true).
+		Insecure(insecure).
 		Build()
 }
 
 // newDVForCR returns a VM specification of the fetched oVirt VM
```
masayag: method description could be more precise, i.e. 'newDVForCR returns the data-volume specifications for the target VM based on oVirt VM'

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
-func newDVForCR(vm *ovirtsdk.Vm, secret *corev1.Secret) []cdiv1.DataVolume {
-	quantity, _ := resource.ParseQuantity("1Gi")
-	diskAttachments, _ := vm.DiskAttachments()
+func newDVForCR(diskAttachments *ovirtsdk.DiskAttachmentSlice, secret *corev1.Secret) []cdiv1.DataVolume {
+	quantity, _ := resource.ParseQuantity("1Gi") // FIXME:
```
masayag: we'll probably need units converter for this.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
-		dvs = append(dvs,
-			cdiv1.DataVolume{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      disk.MustId(), // FIXME:
-					Namespace: "default",     // FIXME:
-					Labels: map[string]string{ // FIXME:
-						"origin": "oVirt",
+		dvs[i] = cdiv1.DataVolume{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "cdi.kubevirt.io/v1alpha1",
```
masayag: pls extract `"cdi.kubevirt.io/v1alpha1"` to a constant

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      diskAttachment.MustId(), // FIXME:
+				Namespace: "default",               // FIXME:
+				Labels: map[string]string{ // FIXME:
+					"origin": "oVirt",
```
masayag: pls extact `"oVirt"` to a constant

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+				},
+			},
+			Spec: cdiv1.DataVolumeSpec{
+				Source: cdiv1.DataVolumeSource{
+					Imageio: &cdiv1.DataVolumeSourceImageIO{
+						URL:           string(secret.Data["apiUrl"]),
+						DiskID:        disk.MustId(),
+						SecretRef:     "ovirt-key", // FIXME, should be created dynamically
```
masayag, line 244: We should probably take care to delete the secret once the data-volume creation is complete.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+		}
 	}
 	return dvs
 }
 
 // newVMForCR returns a VM specification of the fetched oVirt VM
 func newVMForCR(vm *ovirtsdk.Vm, dvs []cdiv1.DataVolume, con *ovirtsdk.Connection) *kubevirtv1.VirtualMachine {
+	// Labels definition:
 	labels := map[string]string{
 		"origin": "oVirt",
```
masayag: pls reuse constant from previous comment for `oVirt`

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -118,146 +127,170 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+		}
+	}
+	running := false
 	return &kubevirtv1.VirtualMachine{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      vm.MustName(), // FIXME:
 			Namespace: "default",     // FIXME:
 			Labels:    labels,
 		},
 		Spec: kubevirtv1.VirtualMachineSpec{
```
masayag, line 321: We should also take care of the boot sequence.

masayag (changes requested):

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Right, thanks.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
machacekondra: Added

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: The question is why do we need to watch the vms? What are we looking for to be notified about? We are interested in progress of the disks and their status.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: With admission [hook](https://github.com/kubevirt/vm-import-operator/blob/master/pkg/webhook/admission/vm-import-validation.go) we may need to fetch it twice or find a way to share source vm definition.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: We could use labels or annotations depending on whether we would query for it or not.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: We would need to implement naming normalization so ovirt names could be [used](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names) in the cluster cluster.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: I think we can explore [quantity](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go)  here

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: From UI perspective Marek wrote GC [controller](https://github.com/oVirt/v2v-conversion-host/blob/master/kubevirt-vmware/pkg/controller/garbage_collector/v2vvmware_garbagecollector.go#L162) to handle that. We need to consider that some/most of the flows will be handled by it.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Correct. Ovirt can have only one bootable disk, so I will set this disk booOrder to 0, and I won't set any bootOrder to other disks.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: ok

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: done

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: done

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: done

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Right, done.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Right, we don't define target namespace anywhere.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: We would need ovirtConnection there as well, because of FollowLink call.

pkliczewski:
@masayag What do you think?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
masayag: My only thought is if we'd need to start the VM and monitor the progress of it.
We can say that the only concern is to to try and start the VM and that's it. without monitoring how that ends.

pkliczewski: I would change this question a bit: do we need to perform an action if the vm fails to start? It brings broader question how to verify whether the import was successful.

pkliczewski (approved):
Let's improve this in follow-up PRs.

masayag (approved):

PR #6, opened 2020-03-09
NIC validation

jakub-dzon:
First proper validations - performed against NIC. The result of the validation will be interpreted in the caller and transformed into `k8s.io/apimachinery/pkg/apis/meta/v1.StatusCause` or logged.
This PR shows the general direction of validation implementation.

### DIFF
--- a/pkg/webhook/admission/validator.go
+++ b/pkg/webhook/admission/validator.go
@@ -55,7 +55,7 @@ func validateNic(nic *ovirtsdk.Nic) []ValidationFailure {
 
 	//TODO: Networking rule #1
 
-	if failure, valid := isValidatNicInterfaceModel(nic, nicID); !valid {
+	if failure, valid := isValidNicInterfaceModel(nic, nicID); !valid {
 		results = append(results, failure)
 	}
 	if failure, valid := isValidNicOnBoot(nic, nicID); !valid {
@@ -69,7 +69,7 @@ func validateNic(nic *ovirtsdk.Nic) []ValidationFailure {
 		if failure, valid := isValidVnicPortMirroring(vNicProfile, nicID); !valid {
 			results = append(results, failure)
 		}
-		if failure, valid := isValidVnicProfileMigratable(vNicProfile, nicID); !valid {
+		if failure, valid := isValidVnicProfilePassThrough(vNicProfile, nicID); !valid {
 			results = append(results, failure)
 		}
 		if failure, valid := isValidVnicProfileCustomProperties(vNicProfile, nicID); !valid {
@@ -86,12 +86,12 @@ func validateNic(nic *ovirtsdk.Nic) []ValidationFailure {
 	return results
 }
 
-func isValidatNicInterfaceModel(nic *ovirtsdk.Nic, nicID string) (ValidationFailure, bool) {
+func isValidNicInterfaceModel(nic *ovirtsdk.Nic, nicID string) (ValidationFailure, bool) {
 	iFace, ok := nic.Interface()
 	if _, found := validInterfaceDeviceModels[string(iFace)]; !ok || !found {
 		return ValidationFailure{
 			ID:      NicInterfaceCheckID,
-			Message: fmt.Sprintf("interface %v uses model %s that is not supported.", nicID, iFace),
+			Message: fmt.Sprintf("interface %s uses model %s that is not supported.", nicID, iFace),
 		}, false
 	}
 	return ValidationFailure{}, true
@@ -101,7 +101,7 @@ func isValidNicOnBoot(nic *ovirtsdk.Nic, nicID string) (ValidationFailure, bool)
 	if onBoot, _ := nic.OnBoot(); !onBoot {
 		return ValidationFailure{
 			ID:      NicOnBootID,
-			Message: fmt.Sprintf("interface %v is not enabled on boot.", nicID),
+			Message: fmt.Sprintf("interface %s is not enabled on boot.", nicID),
 		}, false
 	}
 
@@ -112,7 +112,7 @@ func isValidNicPlugged(nic *ovirtsdk.Nic, nicID string) (ValidationFailure, bool
 	if plugged, _ := nic.Plugged(); !plugged {
 		return ValidationFailure{
 			ID:      NicPluggedID,
-			Message: fmt.Sprintf("interface %v is unplugged.", nicID),
+			Message: fmt.Sprintf("interface %s is unplugged.", nicID),
 		}, false
 	}
 	return ValidationFailure{}, true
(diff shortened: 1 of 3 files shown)


jakub-dzon:
@masayag, @machacekondra, @pkliczewski - could you, please, have a look?

pkg/webhook/admission/validator.go
```diff
@@ -0,0 +1,169 @@
...
+
+	return results
+}
+
+func isValidatNicInterfaceModel(nic *ovirtsdk.Nic, nicID string) (ValidationFailure, bool) {
+	iFace, ok := nic.Interface()
+	if _, found := validInterfaceDeviceModels[string(iFace)]; !ok || !found {
+		return ValidationFailure{
+			ID:      NicInterfaceCheckID,
+			Message: fmt.Sprintf("interface %v uses model %s that is not supported.", nicID, iFace),
```
pkliczewski: Is there correct order of the parameters? `nicID` is a string and `iFace` is `NicInterface` type. Based on [fmt](https://golang.org/pkg/fmt/) you can use %s for strings. This applies to entire file.

pkg/webhook/admission/validator.go
```diff
@@ -0,0 +1,169 @@
...
+	// NicVNicNetworkFilterID defines an ID of a vnic_profile.networ_filter presence check
+	NicVNicNetworkFilterID = CheckID("nic.vnic_profile.network_filter")
+	// NicVNicQosID defines an ID of a vnic_profile.qos presence check
+	NicVNicQosID = CheckID("nic.vnic_profile.qos")
+)
+
+var validInterfaceDeviceModels = map[string]*struct{}{"e1000": nil, "rtl8139": nil, "virtio": nil}
+
+// CheckID identifies validation check for Virtual Machine Import
+type CheckID string
```
pkliczewski, line 33: How are we going to differentiate which `ValidationFailure` should be logged or we need to warn or block the import process? We could add validation level field which would allow us to interpret the slice with failures.

jakub-dzon: I saw two options there: one was what you suggest, the other (chosen) was to separate the concerns of "action" assignment and detection of suspicious configuration elements. I decided to allow the client of the validation methods to decide how to treat each of the failures based on client-owned assignment of `CheckID` to action (i.e. map[CheckID]Action). Despite the other solution being simpler in implementation I feel that the 'level' does not belong to the `ValidationFailure`.  
Do you think the other approach would be better/clearer/cleaner?

pkliczewski: We know that actions when specific rules will change over time as features supported by Kubevirt will grow. We need to have this mapping somewhere and update it over time. My idea was to have the mapping together with rule definition and separate action implementation. My intention is simplicity so I would love to see what you have in mind in the code.

masayag: +1 for simplicity. I'd add a category for each check as mentioned elsewhere (audit / warn / block) and fail the validation on 'block' violations.

pkg/webhook/admission/validator.go
```diff
@@ -0,0 +1,169 @@
...
+		results = append(results, failure)
+	}
+	if failure, valid := isValidNicOnBoot(nic, nicID); !valid {
+		results = append(results, failure)
+	}
+	if failure, valid := isValidNicPlugged(nic, nicID); !valid {
+		results = append(results, failure)
+	}
+
+	if vNicProfile, ok := nic.VnicProfile(); ok {
```
masayag, line 70: IIUC, VnicProfile needs to be fetched by followLink since it is just a reference to the resource and not the entire entity

pkg/webhook/admission/validator.go
```diff
@@ -0,0 +1,169 @@
...
+
+	return results
+}
+
+func isValidNicInterfaceModel(nic *ovirtsdk.Nic, nicID string) (ValidationFailure, bool) {
+	iFace, ok := nic.Interface()
+	if _, found := validInterfaceDeviceModels[string(iFace)]; !ok || !found {
+		return ValidationFailure{
+			ID:      NicInterfaceCheckID,
+			Message: fmt.Sprintf("interface %s uses model %s that is not supported.", nicID, iFace),
```
masayag: worth to list here the supported models to let the user know to which type the nic could be changed.

pkg/webhook/admission/validator.go
```diff
@@ -0,0 +1,169 @@
...
+func isValidVnicProfileCustomProperties(vNicProfile *ovirtsdk.VnicProfile, nicID string) (ValidationFailure, bool) {
+	if cp, ok := vNicProfile.CustomProperties(); ok && len(cp.Slice()) > 0 {
+		return ValidationFailure{
+			ID:      NicVNicCustomPropertiesID,
+			Message: fmt.Sprintf("interface %s uses profile with custom properties: %v.", nicID, cp),
+		}, false
+	}
+	return ValidationFailure{}, true
+}
+func isValidVnicProfileNetworFilter(vNicProfile *ovirtsdk.VnicProfile, nicID string) (ValidationFailure, bool) {
```
masayag, line 155: please add space before function.

pkg/webhook/admission/validator.go
```diff
@@ -0,0 +1,169 @@
...
+	if nf, ok := vNicProfile.NetworkFilter(); ok {
+		return ValidationFailure{
+			ID:      NicVNicNetworkFilterID,
+			Message: fmt.Sprintf("interface %s uses profile with a network filter: %v.", nicID, nf),
+		}, false
+	}
+	return ValidationFailure{}, true
+}
+func isValidVnicProfileQos(vNicProfile *ovirtsdk.VnicProfile, nicID string) (ValidationFailure, bool) {
```
masayag, line 165: please add space before function.

pkg/webhook/admission/validator.go
```diff
@@ -0,0 +1,169 @@
+package admission
```
masayag, line 1: This file will grow beyond reasonable size. I'd rename it to be specific for nic.
Then add another file for disk, affinity and so on.

masayag (changes requested):

pkg/webhook/admission/validator.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+		if failure, valid := isValidVnicProfileQos(vNicProfile, nicID); !valid {
+			results = append(results, failure)
+		}
+	}
+
+	return results
+}
+
+func isValidNicInterfaceModel(nic *ovirtsdk.Nic, nicID string) (ValidationFailure, bool) {
```
masayag, line 91: Regarding this, we could rely on the resource mappings to define the correct mapping between the model which isn't supported by kubevirt to one which is supported.

pkg/webhook/admission/validator.go
jakub-dzon: :+1: I already have my storage validation implemented this way on a branch.

pkg/webhook/admission/validator.go
jakub-dzon: Good one! I assume at that level that I've got everything I need - the client is expected to provide all necessary data - will pay attention that at that level.

pkg/webhook/admission/validator.go
jakub-dzon: I was thinking about something like 
```
type action int

const (
	log   = 0
	warn  = 1
	block = 2
)

var checkToAction = map[CheckID]action{
	NicInterfaceCheckID:       block,
	NicOnBootID:               log,
	NicPluggedID:              warn,
	NicVNicPassThroughID:      block,
	NicVNicPortMirroringID:    warn,
	NicVNicCustomPropertiesID: warn,
	NicVNicNetworkFilterID:    warn,
	NicVNicQosID:              log,
}
```
in vm-import-validation.go
and then instead of getting the action from the `ValidationFailure` object, I would look it up in the  `checkToAction` whil processing the slice of failures.
I'm fine with having it in the `ValidationFailure`, though.

pkg/webhook/admission/validator.go
jakub-dzon: Will do.

pkg/webhook/admission/validator.go
jakub-dzon: Yes, ideally there should be only one source of truth.

pkg/webhook/admission/validator.go
jakub-dzon: It looks like this: https://github.com/jakub-dzon/vm-import-operator/tree/storage_rules/pkg/webhook/admission. If that's OK, I will keep the split-related changes there.

pkliczewski (approved):

masayag (approved):

pkg/webhook/admission/validator.go
jakub-dzon: The checkToAction map will be implemented in a different PR.

PR #8, opened 2020-03-11
Storage rules

jakub-dzon:
This PR:
-  defines validation rules for storage;
-  puts different sets of validators in separate files (networking, storage);
-  introduces `CheckID`:`action` map that will be used in the hook to decide what strategy should be executed to deal with specific `ValidationFailure`. Strategies to be implemented in other PR.

### DIFF
--- a/pkg/webhook/admission/nic-validator.go
+++ b/pkg/webhook/admission/nic-validator.go
@@ -6,39 +6,10 @@ import (
 	ovirtsdk "github.com/ovirt/go-ovirt"
 )
 
-const (
-	// NicInterfaceCheckID defines an ID of a NIC interface model check
-	NicInterfaceCheckID = CheckID("nic.interface")
-	// NicOnBootID defines an ID of a NIC on_boot == fales check
-	NicOnBootID = CheckID("nic.on_boot")
-	// NicPluggedID defines an ID of a NIC plugged == false check
-	NicPluggedID = CheckID("nic.plugged")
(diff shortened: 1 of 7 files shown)


pkg/webhook/admission/nic-validator.go
```diff
@@ -6,40 +6,11 @@ import (
...
-	NicVNicPassThroughID = CheckID("nic.vnic_profile.pass_through")
-	// NicVNicCustomPropertiesID defines an ID of a vnic_profile.custom_properties presence check
-	NicVNicCustomPropertiesID = CheckID("nic.vnic_profile.custom_properties")
-	// NicVNicNetworkFilterID defines an ID of a vnic_profile.networ_filter presence check
-	NicVNicNetworkFilterID = CheckID("nic.vnic_profile.network_filter")
-	// NicVNicQosID defines an ID of a vnic_profile.qos presence check
-	NicVNicQosID = CheckID("nic.vnic_profile.qos")
-)
-
 // TODO: move to shared package with mapping definitions
```
pkliczewski, line 9: Is this TODO still valid?

jakub-dzon: Yes - I envisioned this and storage maps to be used both here - for validation, and in reconciliation - for mapping from oVirt to kubevirt values.

machacekondra (approved):

masayag (approved):

PR #9, opened 2020-03-11
Add ResourceMapping CRD

masayag:
OvirtResourceMapping CRD is used to describe the mapping of oVirt resources
to KubeVirt.

oVirt resources can be identified by name (if unique) or by ID.


README.md
```diff
@@ -31,30 +32,33 @@ stringData:
...
 ```
 
-## Create a config map that defines the resource mappings from oVirt to KubeVirt:
+## Create oVirt resource mappings that defines how oVirt resources are mapped to kubevirt:
 ```bash
 cat <<EOF | kubectl create -f -
-apiVersion: v1
-kind: ConfigMap
+apiVersion: v2v.kubevirt.io/v1alpha1
+kind: OvirtResourceMapping
```
pkliczewski: What do you think about source infrastructure agnostic name? We could reuse this mapping CR for other imports.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
```diff
@@ -70,8 +61,7 @@ type VirtualMachineImportOvirtSourceVMSpec struct {
 // +k8s:openapi-gen=true
 // +optional
 type VirtualMachineImportOvirtSourceVMClusterSpec struct {
-	// +optional
-	ID *string `json:"id,omitempty"`
+	ID string `json:"id"`
```
pkliczewski: Why do we want to make `ID` as required?

README.md
masayag: I'm not sure about that. I'd prefer it be provider's specific since other providers might identify their resources differently and even in ovirt not all resources are identified the same way.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
masayag: If the ID is required, there is no need for name field at all.
Providing ID is the recommended way, however, it isn't human readable, hence added an option to specify the name instead.

machacekondra: But in this case user must specify the ID, so we don't have to have the name field at all.

pkg/apis/v2v/v1alpha1/ovirtresourcemapping_types.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// +optional
+	ID *string `json:"id,omitempty"`
+}
+
+// ResourceMappingItem defines the mapping of a single resource from ovirt to kubevirt
+// +k8s:openapi-gen=true
+type ResourceMappingItem struct {
+	Source Source `json:"source"`
+	Target string `json:"target"`
```
machacekondra: Would it be always sufficient this to be string? Wouldn't it be better to make it object, so it can be enhanced in future?

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
jakub-dzon: IMO, if we have he ID required, we should resign from the name field. Either way - name or ID the user will probably copy&paste the value, so I would just leave the id.

pkg/apis/v2v/v1alpha1/ovirtresourcemapping_types.go
jakub-dzon: :+1: Eventually we may need to map more complicated structures.

README.md
masayag: based on planning to support other provider types and in order to comply with multi-provider support in other resources (vmimport, secret data) the resource is changed to ResourceMapping and added ResourceMapping.spec.ovirt for ovirt's mappings.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
masayag: just reading it, i realized reading Piotr's the other way around :-)
The Id was intended to remain optional. Should be fine in next update.

pkg/apis/v2v/v1alpha1/resourcemapping_types.go
```diff
@@ -0,0 +1,83 @@
...
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// +optional
+	OvirtMappings *OvirtMappings `json:"ovirt,omitempty"`
+}
+
+// OvirtMappings defines the mappings of ovirt resources to kubevirt
+// +k8s:openapi-gen=true
+type OvirtMappings struct {
```
jakub-dzon, line 23: Can we have a more generic type for the mappings? Will mapping for VMware be different?
Having `type Mappings`, not bound to any provider we can implement the provider "plug-in" API in a more generic way in some of its parts - for example in validation, where the mapping needs to be verified we can pass just the `Mapping` instead of its properties separately.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
```diff
@@ -83,36 +80,65 @@ type VirtualMachineImportStatus struct {
...
+)
+
+// ProcessingConditionReason defines the reasons for the Processing condition of VM import
+// +k8s:openapi-gen=true
+type ProcessingConditionReason string
+
+// These are valid reasons for the Processing conditions of VM import.
+const (
+	// Validating represents the validation of the mapping rules and eligibility of source VM for import
+	Validating ProcessingConditionReason = "Validating"
```
jakub-dzon: Having these allowed reasons and condition types should the validation failure be represented as follows? 
- block: type: "Processing", reason: "Validating", status: "False"
- warn:  type: "Processing", reason: "Validating", status: "True"

pkg/apis/v2v/v1alpha1/resourcemapping_types.go
masayag: For VMWare we'll have VMWareMappings struct. Then we'll add 'oneOf' constraint to the schema to validate only one type of mappings is provided.

See https://docs.google.com/document/d/1uz_E2Jydb1dA_1j8N0M-j-PKa-c6gPzu6qpZdZPptu8/edit?disco=AAAAJM9TOZU

So we have a sections per provider for:
* secret.data
* vmimport.spec.source
* resourcemapping.spec
that should allow to support other providers within the same resources.

VMWare might define the mappings differently. Internally we can define an interface to gain polymorphic-like behavior, but it should influence the representation of the data.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
masayag: This validation reason includes any possible cause for validation failures:
* Secret not found
* Mappings resource not found
* Failure to connect to source provider
* Failure to to fetch VM details
* Failure to prepare mapping rules
* Violation of mapping rules

Therefore I think that validation is entitled for its own condition in order to provide a meaningful feedback about the cause of the failure above.

Would you see a value if separating the validating step from being tracked by the 'Processing' condition ?
And we'll have two conditions for monitoring the progress:
'Validating' - verifies the eligibility of the input for the import process.
'Processing' - once the inputs are verified, monitors the processing of the vm import process

Then 'Validating' condition will a similar behavior as you suggested:
* block: type: "Validating", reason: "MappingRulesViolated", status: "False"

But in order to be able and record the warning, we should add a dedicated condition for it, else, it will be overridden by next update to that condition.

You should add a condition 'MappingCheck' with reasons:
- MappingRulesViolated - status 'False' // message should contain the violations 
- MappingRulesHaveWarnings - status 'True' // message should contain the warnings 
- MappingRulesReady - status 'True'

Feel free to propose other naming.

pkg/apis/v2v/v1alpha1/resourcemapping_types.go
jakub-dzon: OK, sounds good :+1:

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
jakub-dzon: Yes, I think we should have separate condition types for reporting validation problems.

jakub-dzon: I think type "Validating" and reasons you proposed look good.

README.md
```diff
@@ -23,38 +24,35 @@ metadata:
...
-      - source: red # maps of ovirt logic network to network attachment definition
-        target: xyz
-        type: bridge
-      - source: ovirtmgmt
-        target: pod
+  name: example-resourcemappings
+  namespace: example-ns
+spec:
+  ovirt:
+    networkMappings:
```
pkliczewski, line 47: I wonder whether it is enough to identify storagedomain or network by name.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
```diff
@@ -83,36 +80,65 @@ type VirtualMachineImportStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
-	TargetVirtualMachineName string                          `json:"targetVirtualMachineName"`
-	Conditions               []VirtualMachineImportCondition `json:"conditions"`
-	State                    VirtualMachineImportState       `json:"state"`
-	DataVolumes              []DataVolumeItem                `json:"dataVolumes"`
+	TargetVMName string                          `json:"targetVmName"`
```
pkliczewski, line 83: I would add validation warnings section here. Validation rules may generate good number of them.

README.md
masayag: The definition of the source is based on https://github.com/kubevirt/vm-import-operator/pull/9/files#diff-7b282a764b7b115b17d31e0cdae066caR33 which accepts also `ID`.

If we have the origin VM, we can get its cluster and the data-center ID and using the resource name (network name or storage domain name) to uniquely identify it.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
masayag: Should it be as a section in the status  or as a message of the mapping validation conditions ?

pkliczewski: I am not sure which section would be the best. I am thinking about the warnings that will be outcome of source vm validation rules that Jakub is working on.

jakub-dzon: So far we've settled on putting these warning messages in a condition of a dedicated ("Validating") type with following logic:
if errors are present, they are put in the message else if there are warnings then they are used, else "All good"-like message is used. Do you think it should work differently?

pkliczewski: I just wanted to make sure we have place for them in the spec. Condition is fine.

masayag: See https://github.com/kubevirt/vm-import-operator/pull/9/files#diff-1a98d2071d9686a1e97f4596c788f860R101
The `MappingRulesChecking` will report the violations by updating the 'message' field and setting the proper value for the reason from the proposed values:
https://github.com/kubevirt/vm-import-operator/pull/9/files#diff-1a98d2071d9686a1e97f4596c788f860R162

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
```diff
@@ -83,34 +80,110 @@ type VirtualMachineImportStatus struct {
...
+	UnreachableProvider ValidatingConditionReason = "UnreachableProvider"
+
+	// SourceVmNotFound represents the nonexistence of the source VM
+	SourceVMNotFound ValidatingConditionReason = "SourceVMNotFound"
+
+	// IncompleteMappingRules represents the inability to prepare the mapping rules
+	IncompleteMappingRules ValidatingConditionReason = "IncompleteMappingRules"
+
+	// MappingRulesViolated represents the violation of mapping rules
+	MappingRulesViolated ValidatingConditionReason = "MappingRulesViolated"
```
jakub-dzon: What is the difference between `Validating/MappingRulesViolated` and `MappingRulesChecking.MappingRulesCheckingFailed`?

deploy/crds/v2v_v1alpha1_virtualmachineimport_crd.yaml
```diff
@@ -32,9 +32,66 @@ spec:
...
+              description: 'TargetVmName defines the name of the target VM to be created'
+              type: string
+            startVm:
+              description: 'StartVm defines if the VM should be started at the end of the import process'
+              type: boolean
+            source:
+              description: 'VirtualMachineImportSourceSpec defines the definition of the source provider and mapping resources'
+              properties:
+                ovirt:
+                  description: 'VirtualMachineImportOvirtSourceSpec defines the definition of the VM in oVirt and the credentials to oVirt'
```
machacekondra: ` defines the definition of the VM in oVirt and the credentials to oVirt`
the credentials is not true anymore

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
masayag: `Validating/MappingRulesViolated` - the reason for failing the validation of import vm process is a blocking violation of mapping rules. That will cause the entire VMImport process to fail. the message in this context is concise.
In this case, we'll have also the condition `MappingRulesChecking` set to `False` with reason `MappingRulesCheckingFailed` and the message should list all of the violations.

If mapping rules check produces only warnings, `MappingRulesChecking` is set to 'True`, the reason will be `MappingRulesCheckingReportedWarnings` and the message of that condition should list all of the warnings.

That will guarantee that  even if `Validating` is true, the warnings remain part of the status, kept on `MappingRulesChecking` condition.

jakub-dzon: I understand, thanks.
In that case it seems to me that `Validating/MappingRulesViolated` is superfluous if the failure listing will be in the `MappingRulesChecking` and its status will also be `False`.
What do you think about removing that reason from `Validating`? Then  `Validating` will represent "technical issues or user input checks" and `MappingRulesChecking` will focus on the rules.

README.md
```diff
@@ -23,38 +24,35 @@ metadata:
   name: my-secret-with-ovirt-credentials
 type: Opaque
 stringData:
-  config.yaml: |-
+  ovirt: |-
     apiUrl: "https://ovirt-engine.example.com:8443/ovirt-engine/api/"
     username: admin@internal # provided in the format of username@domain
     password: 123456
     insecure: true
```
pkliczewski: I think that setting this flag is not needed. If the `ca` is not provided we should set insecure to `true`.

masayag: @machacekondra mentioned that without specifying  `insecure`, the ovirt-client will attempt using system CA. For insecure, it is necessary to provide it.

@machacekondra pls correct me if i'm wrong

pkliczewski: We need to set it. My point is why user needs to provide. We can figure it out based on ca availability and hide this detail from a user.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
masayag: Sure, we can do that the same way as we excluded the 'validating' from the 'processing' condition.

README.md
machacekondra: Well the logic is:
1) if user provide CA use provided CA
2) if user don't provide CA use system store
3) if user use insecure = true, use insecure connection.

Meaning, that user must explicitly say to use insecure connection, and don't use it by default.

pkliczewski: I think in a container we won't be able to use second point.

jakub-dzon (approved):

pkliczewski (approved):

machacekondra (approved):

PR #10, opened 2020-03-13
VM rules

jakub-dzon:
This PR adds VM validation rules.


pkg/webhook/admission/definitions.go
```diff
@@ -43,8 +43,70 @@ const (
...
+	VMGraphicConsolesID = CheckID("vm.graphic_consoles.protocol")
+	// VMHostDevicesID defines an ID of a vm.host_devices check
+	VMHostDevicesID = CheckID("vm.host_devices")
+	// VMReportedDevicesID defines an ID of a vm.reported_devices check
+	VMReportedDevicesID = CheckID("vm.reported_devices")
+	// VMQuotaID defines an ID of a vm.quota check
+	VMQuotaID = CheckID("vm.quota")
+	// VMWatchdogsID defines an ID of a vm.watchdogs
+	VMWatchdogsID = CheckID("vm.watchdogs")
+	// VMCdromsID defines ID of a vm.cdroms storage domain type check
```
pkliczewski: s/ID/an ID here and the last `ID`

pkg/webhook/admission/vm-validator.go
```diff
@@ -0,0 +1,482 @@
...
+		}
+		if failure, valid := isValidBiosType(bios); !valid {
+			results = append(results, failure)
+		}
+	}
+	return results
+}
+
+func isValidBootMenu(bios *ovirtsdk.Bios) (ValidationFailure, bool) {
+
```
pkliczewski: This empty line seems not to be consistent with other functions.

pkg/webhook/admission/vm-validator.go
```diff
@@ -0,0 +1,482 @@
...
+	}
+	return ValidationFailure{}, true
+}
+
+func isValidRandomNumberGeneratorSource(vm *ovirtsdk.Vm) (ValidationFailure, bool) {
+	if rng, ok := vm.RngDevice(); ok {
+		if src, ok := rng.Source(); ok && src != "urandom" {
+			return ValidationFailure{
+				ID:      VMRngDeviceSourceID,
+				Message: fmt.Sprintf("VM has illegal random number generator device source set: %v. Supported value: 'urandom'", src),
```
pkliczewski: Is it illegal or unsupported?

jakub-dzon: It cannot be mapped in any way - CNV will always use host's urandom if we specify rng option in the spec. Probably "unsupported" would be better.

pkliczewski (approved):

machacekondra (approved):

masayag (approved):

PR #11, opened 2020-03-16
oVirt VM validation

jakub-dzon:
This PR:
 - introduces `providers` package in preparation for future support for providers other than oVirt;
 - moves validators to separate sub-package;
 - introduces admission webhook-agnostic VM validation API;


pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -0,0 +1,270 @@
...
+type OvirtClient interface {
+	GetVM(connectionSettings *ConnectionSettings, id *string, name *string, cluster *string) (*ovirtsdk.Vm, error)
+}
+
+// ConnectionSettings wrap information required to make oVirt API connection
+type ConnectionSettings struct {
+	URL      string
+	Username string
+	Password string
+	Insecure bool
```
pkliczewski: We need a field for CA content to make the connection secure.

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func (client *RichOvirtClient) fetchVM(con *ovirtsdk.Connection, id *string, name *string, cluster *string) (*ovirtsdk.Vm, error) {
+	// Id of the VM specified:
+	if id != nil {
+		response, err := con.SystemService().VmsService().VmService(*id).Get().Send()
+		if err != nil {
+			return nil, err
+		}
+		vm, _ := response.Vm()
```
pkliczewski: What if the vm is not there?

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -0,0 +1,270 @@
...
+	vms, _ := response.Vms()
+	if len(vms.Slice()) > 0 {
+		return vms.Slice()[0], nil
+	}
+	return nil, fmt.Errorf("Virtual machine %v not found", *name)
+}
+
+func (client *RichOvirtClient) populateHostDevices(con *ovirtsdk.Connection, vm *ovirtsdk.Vm) error {
+	if hdcs, ok := vm.HostDevices(); ok {
+		if _, ok := hdcs.Href(); ok {
```
pkliczewski: What is the value of `Href` check here? It is done as second check in `FollowLink` function.

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -0,0 +1,270 @@
...
+// ConnectionSettings wrap information required to make oVirt API connection
+type ConnectionSettings struct {
+	URL      string
+	Username string
+	Password string
+	Insecure bool
+}
+
+// RichOvirtClient is responsilbe for retrieving VM data from oVirt API
+type RichOvirtClient struct {
```
pkliczewski: Why not to keep the connection in the struct?

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// GetVM rertrieves oVirt VM data for given id or name and cluster. VM will have certain links followed and updated.
+func (client *RichOvirtClient) GetVM(cs *ConnectionSettings, id *string, name *string, cluster *string) (*ovirtsdk.Vm, error) {
+	con, err := client.createConnection(cs.URL, cs.Username, cs.Password, cs.Insecure)
```
pkliczewski: please add defer con.Close(). We do not want to leak open connections.

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+	return nil
+}
+
+func (client *RichOvirtClient) createConnection(apiURL string, username string, password string, insecure bool) (*ovirtsdk.Connection, error) {
+	connection, err := ovirtsdk.NewConnectionBuilder().
+		URL(apiURL).
+		Username(username).
+		Password(password).
+		Insecure(insecure).
```
pkliczewski: if ca provided we need to make if secure.

pkg/providers/ovirt/validation/vm-import-validation.go
```diff
@@ -0,0 +1,124 @@
...
+	valid := true
+	var propagatedFailures []validators.ValidationFailure
+	for _, failure := range failures {
+		switch checkToAction[failure.ID] {
+		case log:
+			klog.Infof("Validation information: %v", failure)
+		case warn:
+			propagatedFailures = append(propagatedFailures, failure)
+		case block:
+			valid = false
```
pkliczewski, line 140: What do you think that if we see any block action we return new error and stop processing?

pkg/providers/ovirt/validation/vm-import-validation.go
```diff
@@ -0,0 +1,124 @@
...
+
+// NewVirtualMachineImportValidator creates ready-to-use NewVirtualMachineImportValidator
+func NewVirtualMachineImportValidator() VirtualMachineImportValidator {
+	return VirtualMachineImportValidator{
+		Validator: &validators.ValidatorWrapper{},
+	}
+}
+
+// Validate validates whether VM described in VirtualMachineImport can be imported
+func (validator *VirtualMachineImportValidator) Validate(vm *ovirtsdk.Vm) ([]validators.ValidationFailure, bool) {
```
pkliczewski: Where do we need `[]ValidationFailure` slice out side of this method?

jakub-dzon: In code that will add the failure information to the CR - probably in the reconciliation loop.

pkg/providers/ovirt/validation/vm-import-validation.go
jakub-dzon: I wouldn't want to use error for reporting that. I see `valid==false` result as something that is an expected, correct behaviour of the application and not something abnormal that should be an error.

pkg/providers/ovirt/client/ovirt-client.go
jakub-dzon: :+1:

pkg/providers/ovirt/client/ovirt-client.go
jakub-dzon: `FollowLink` returns error due to several reasons and that's the only one I want to ignore without relying on error messages.

pkg/providers/ovirt/client/ovirt-client.go
jakub-dzon: In that case, I think, we would have to expose con.Close() as another function of that API and add probably some logic for checking if it's connected or not.
I see more value in moving `ConnectionSettings` from the method arguments to the struct and connecting/disconnecting in the retrieval logic as it is. What do you think?

pkg/providers/ovirt/client/ovirt-client.go
jakub-dzon: :+1:

pkg/providers/ovirt/client/ovirt-client.go
pkliczewski: You still return error if something is not right. Not sure why to handle something twice :thinking:

pkg/providers/ovirt/client/ovirt-client.go
pkliczewski: Reconnecting takes time like handshaking and creating http session. Please check any examples from [here](https://github.com/oVirt/ovirt-engine-sdk-go/tree/master/sdk/examples) to see how complex is Close().

pkg/providers/ovirt/validation/vm-import-validation.go
pkliczewski: It seems like `processFailures` is not really processing `ValidationFailure`s. We log and filter log action and return everything else.

pkg/providers/ovirt/validation/vm-import-validation.go
pkliczewski: Why having an error is so bad? Once we know we need to block we do not need to do anything else.

pkg/providers/ovirt/client/ovirt-client.go
jakub-dzon: No, if there is no href, I don't follow the link. If I followed, the sdk method would return an error, which I would have to dispatch based on the message - to ignore no-href error and propagate others.

pkg/providers/ovirt/client/ovirt-client.go
jakub-dzon: OK. Is your assumption that this client will have more functionality (like starting, stopping vm, updating name/etc. to flag it's been migrated) and we want to connect only once?

pkg/providers/ovirt/validation/vm-import-validation.go
jakub-dzon: Do you mean - breaking the loop? In that case we would have to return full slice of failures to the caller, which would have decide again which ones need to be put in the CR status.

pkg/providers/ovirt/client/ovirt-client.go
jakub-dzon: I have just pushed the client with struct-bound connection. Is that what you had in mind?

pkg/providers/ovirt/client/ovirt-client.go
pkliczewski: OK, it is up to you. You could save good number of lines if you rely on `FollowLink` to check it.

pkg/providers/ovirt/client/ovirt-client.go
pkliczewski: Yes :+1:

pkg/providers/ovirt/validation/vm-import-validation.go
pkliczewski: yes, I meant breaking the loop (stop processing). Having meaningful error would be enough. We could update vm import object with this message and mark it as failed. If you see any better way to propagate this information feel free to implement.

pkg/providers/ovirt/validation/vm-import-validation.go
jakub-dzon: You're right. To make it really process all the failures I will change the API  from `Validate(*ovirtsdk.vm)` to `Validate(v2vvalpha1.VirtualMachineImport)` and update the CR status in that function for warns and blocks.

pkg/providers/ovirt/client/ovirt-client.go
jakub-dzon: Actually, I would have to decide what to do with errors for no href, as that is not an import-breaker.
The core reason for that href pre-check is that I've gotten an error while trying to retrieve quota for a VM from a real, live cluster and that error should not be a blocker.

pkg/providers/ovirt/validation/vm-import-validation.go
jakub-dzon: We need to communicate/preserve all of the issues blocking the import so the user can fix them all and re-submit the CR without iterating from error to error. I will update the CR in that function and won't return the failures to the caller. In that case returning `error` IMO would be acceptable instead of a `bool`.

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -0,0 +1,289 @@
...
+			}
+			vm.SetNics(nics)
+		}
+	}
+	return nil
+}
+
+func (client *richOvirtClient) populateDiskAttachments(vm *ovirtsdk.Vm) error {
+	if das, ok := vm.DiskAttachments(); ok {
+		if _, ok := das.Href(); ok {
```
machacekondra: This check is actually already part of FollowLink: https://github.com/oVirt/go-ovirt/blob/d30f754823a6f8ccd317a23e3ead24d24cbb03de/connection.go#L117

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -0,0 +1,289 @@
...
+			if err != nil {
+				return err
+			}
+			vm.SetWatchdogs(followed.(*ovirtsdk.WatchdogSlice))
+		}
+	}
+	return nil
+}
+
+func (client *richOvirtClient) populateGrapthicsConsoles(vm *ovirtsdk.Vm) error {
```
machacekondra: s/populateGrapthicsConsoles /populateGraphicsConsoles

jakub-dzon: :man_facepalming:

pkg/providers/ovirt/client/ovirt-client.go
jakub-dzon: Please have a look at my explanation: https://github.com/kubevirt/vm-import-operator/pull/11#discussion_r393285221

pkg/providers/ovirt/client/ovirt-client.go
machacekondra: Most of the object must have the href. What was the specific case that your quota didn't have href? If it really happened it's most probably oVirt bug.

jakub-dzon: Actually, the documentation say it is a link, but in the VM xml I just inspected it is just an element:
`  <quota id="3d1bb0a1-785c-4fbe-8b0b-fdb8bc4ae5d7"/>`.
Docs: http://ovirt.github.io/ovirt-engine-api-model/4.4/#types/vm.

jakub-dzon: Seeing that I am not 100% confident that removing all the other checks will always be safe.

pkg/providers/ovirt/validation/vm-import-validation.go
jakub-dzon: Just pushed the changes to enclose whole validation processing in that package. Please have a look if that's better.

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+	return nil
+}
+
+func connect(apiURL string, username string, password string, caFile string, insecure bool) (*ovirtsdk.Connection, error) {
+	connection, err := ovirtsdk.NewConnectionBuilder().
+		URL(apiURL).
+		Username(username).
+		Password(password).
+		CAFile(caFile).
```
pkliczewski: When we have a ca we do not need to set the insecure flag. Do we expect second secret to be mounted as a file?

jakub-dzon: Yes, the CA file should be made available this way or downloaded by the client and put in /tmp.
For testing purposes I think it's useful to still be able to set the connection to insecure.

pkg/providers/ovirt/client/ovirt-client.go
machacekondra: I think we should trust the API, this is indeed bug, and I've opened the bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1814613. I think we should remove the href checks and add it only for quota with comment explaining why it's there, and link to bug. But feel free to leave as is.

jakub-dzon: Thanks! I will remove all the unnecessary checks.

pkliczewski:
@jakub-dzon I wonder whether it would make sense to split this PR into 2. One with validation and the other with the client. This should make the code to get merged faster. What do you think?

jakub-dzon:
OK, doing it.

pkg/providers/ovirt/validation/validators/vm-validator.go
```diff
@@ -423,7 +423,7 @@ func isValidQuota(vm *ovirtsdk.Vm) (ValidationFailure, bool) {
 	if quota, ok := vm.Quota(); ok {
 		return ValidationFailure{
 			ID:      VMQuotaID,
-			Message: fmt.Sprintf("VM has quota assigned: %v", quota),
+			Message: fmt.Sprintf("VM has quota with ID %v assigned", quota.MustId()),
```
pkliczewski: `MustId` may end up with `panic`. It would be great to use `Id` instead.

pkg/providers/ovirt/validation/validators/vm-validator.go
```diff
@@ -452,7 +452,7 @@ func isValidCdroms(vm *ovirtsdk.Vm) []ValidationFailure {
 				if storageType, ok := getStorageDomainType(file); ok && storageType != "data" {
 					results = append(results, ValidationFailure{
 						ID:      VMCdromsID,
-						Message: fmt.Sprintf("VM uses CD ROM with image not stored in data domain: %v", cdrom),
+						Message: fmt.Sprintf("VM uses CD ROM with image not stored in data domain: %v", cdrom.MustId()),
```
pkliczewski: Let's use `Id` instead.

pkliczewski (approved):

machacekondra (approved):

PR #15, opened 2020-03-20
Improve reconcile

machacekondra:

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -137,7 +137,7 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-	ovirt, err := ovirtclient.NewRitchOvirtClient(
+	ovirt, err := ovirtclient.NewRichOvirtClient(
 		&ovirtclient.ConnectionSettings{
 			URL:      ovirtSecretDataMap["apiUrl"],
 			Username: ovirtSecretDataMap["username"],
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -2,6 +2,7 @@ package virtualmachineimport
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 	"strconv"
 
@@ -137,12 +138,23 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
+	if _, ok := ovirtSecretDataMap["caCert"]; !ok {
+		return reconcile.Result{}, fmt.Errorf("oVirt secret must contain caCert attribute")
+	}
+	if len(ovirtSecretDataMap["caCert"]) == 0 {
+		return reconcile.Result{}, fmt.Errorf("oVirt secret caCert cannot be empty")
+	}
+	caCert, err := base64.StdEncoding.DecodeString(ovirtSecretDataMap["caCert"])
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("Failed to decode oVirt secret caCert to base64 format")
+	}
 	ovirt, err := ovirtclient.NewRichOvirtClient(
 		&ovirtclient.ConnectionSettings{
 			URL:      ovirtSecretDataMap["apiUrl"],
 			Username: ovirtSecretDataMap["username"],
 			Password: ovirtSecretDataMap["password"],
-			CACert:   ovirtSecretDataMap["caCert"],
+			CACert:   caCert,
 		},
 	)
 	if err != nil {
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -204,25 +204,13 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 		}
 
 		// Secret with username/password for the image import:
-		secretObj := &corev1.Secret{}
-		err = r.client.Get(context.TODO(), types.NamespacedName{Name: ovirtSecret, Namespace: vmSpec.Namespace}, secretObj)
-		if err != nil && errors.IsNotFound(err) {
-			dvSecret := createDVSecret(ovirtSecretDataMap, instance)
-			err = r.client.Create(context.TODO(), dvSecret)
-			if err != nil {
-				return reconcile.Result{}, err
-			}
+		if err = ensureDVSecretExists(r.client, instance, vmSpec.Namespace, ovirtSecretDataMap); err != nil {
+			return reconcile.Result{}, err
 		}
 
 		// CM containing CA for the image import:
-		cm := &corev1.ConfigMap{}
-		err = r.client.Get(context.TODO(), types.NamespacedName{Name: ovirtConfigmap, Namespace: vmSpec.Namespace}, cm)
-		if err != nil && errors.IsNotFound(err) {
-			dvCm := createDVConfigmap(ovirtSecretDataMap, instance)
-			err = r.client.Create(context.TODO(), dvCm)
-			if err != nil {
-				return reconcile.Result{}, err
-			}
+		if err = ensureDVConfigmapExists(r.client, instance, vmSpec.Namespace, ovirtSecretDataMap); err != nil {
+			return reconcile.Result{}, err
 		}
 
 		// Import disks:
@@ -265,6 +253,26 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 	reqLogger.Info("Skip reconcile: VM already exists", "VM.Namespace", found.Namespace, "VM.Name", found.Name)
 
 	return reconcile.Result{}, nil
+}
+
+func ensureDVSecretExists(client client.Client, instance *v2vv1alpha1.VirtualMachineImport, namespace string, ovirtSecretDataMap map[string]string) error {
+	secretObj := &corev1.Secret{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: ovirtSecret, Namespace: namespace}, secretObj)
+	if err != nil && errors.IsNotFound(err) {
+		dvSecret := createDVSecret(ovirtSecretDataMap, instance)
+		return client.Create(context.TODO(), dvSecret)
+	}
+	return err
+}
+
+func ensureDVConfigmapExists(client client.Client, instance *v2vv1alpha1.VirtualMachineImport, namespace string, ovirtSecretDataMap map[string]string) error {
+	cm := &corev1.ConfigMap{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: ovirtConfigmap, Namespace: namespace}, cm)
+	if err != nil && errors.IsNotFound(err) {
+		dvCm := createDVConfigmap(ovirtSecretDataMap, instance)
+		return client.Create(context.TODO(), dvCm)
+	}
+	return err
 }
 
 func createDVConfigmap(creds map[string]string, vmImport *v2vv1alpha1.VirtualMachineImport) *corev1.ConfigMap {
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -145,10 +145,10 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 			CACert:   ovirtSecretDataMap["caCert"],
 		},
 	)
-	defer ovirt.Close()
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+	defer ovirt.Close()
 
 	// Fetch oVirt VM:
 	sourceVMID := instance.Spec.Source.Ovirt.VM.ID
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -6,7 +6,9 @@ import (
 	"strconv"
 
 	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
+	ovirtclient "github.com/kubevirt/vm-import-operator/pkg/providers/ovirt/client"
 	ovirtsdk "github.com/ovirt/go-ovirt"
+	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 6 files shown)


pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -121,21 +129,38 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-	ovirtConnection, err := createOvirtConnection(ovirtSecret)
+
+	ovirt, err := ovirtclient.NewRitchOvirtClient(
+		&ovirtclient.ConnectionSettings{
+			URL:      string(ovirtSecret.Data["apiUrl"]),
```
masayag: The ovirt secret should look obtain data from `ovirt` section according to the agreed format of the secret. See example on README: https://github.com/kubevirt/vm-import-operator/blob/master/README.md

I updated the example here: https://github.com/kubevirt/vm-import-operator/pull/16

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -121,21 +129,38 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+	defer ovirt.Close()
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
 	// Fetch oVirt VM:
-	vm, err := fetchOvirtVM(ovirtConnection, &instance.Spec.Source.Ovirt.VM)
+	sourceVMID := instance.Spec.Source.Ovirt.VM.ID
+	sourceVMName := instance.Spec.Source.Ovirt.VM.Name
+	sourceVMCluster := instance.Spec.Source.Ovirt.VM.Cluster.Name
```
masayag: cluster can also be identified by ID

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Thanks, done.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Done

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -88,7 +97,58 @@ func (client *richOvirtClient) GetVM(id *string, name *string, cluster *string)
...
+	return nil
+}
+
+// StopVM stop the VM and wait for the vm to be stopped
+func (client *richOvirtClient) StopVM(id string) error {
+	vmService := client.connection.SystemService().VmsService().VmService(id)
+
+	// Return if VM is already stopped:
+	tmp, _ := vmService.Get().Send()
+	if status, _ := tmp.MustVm().Status(); status == ovirtsdk.VMSTATUS_DOWN {
```
jakub-dzon: There is a chance that the VM with the given ID doesn't exist - `StopVM` might be called without any validation beforehand (i.e. `GetVM` call). `MustVm` can panic.

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -88,7 +97,58 @@ func (client *richOvirtClient) GetVM(id *string, name *string, cluster *string)
...
+		return err
+	}
+
+	// Wait for VM to be stopped
+	c := make(chan bool, 1)
+	go func() {
+		for {
+			time.Sleep(vmPollInterval * time.Second)
+			tmp, _ = vmService.Get().Send()
+			if status, _ := tmp.MustVm().Status(); status == ovirtsdk.VMSTATUS_DOWN {
```
jakub-dzon: Is it possible the VM disappears (i.e. someone deletes it on the oVirt side)? If yes, I think that we can panic here.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -146,23 +179,66 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+
+		// Create kubevirt VM from oVirt VM:
 		reqLogger.Info("Creating a new VM", "VM.Namespace", vmSpec.Namespace, "VM.Name", vmSpec.Name)
+		err = r.client.Create(context.TODO(), vmSpec)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		// Secret with username/password for the image import:
+		secretObj := &corev1.Secret{}
```
jakub-dzon: What do you think about extracting the whole block to a separate function (`ensureDVSecretExists`)?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -146,23 +179,66 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+		if err != nil && errors.IsNotFound(err) {
+			dvSecret := createDVSecret(ovirtSecretDataMap, instance)
+			err = r.client.Create(context.TODO(), dvSecret)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+
+		// CM containing CA for the image import:
+		cm := &corev1.ConfigMap{}
```
jakub-dzon: What do you think about extracting the whole block to a separate function (`ensureDVConfigmapExists`)?

pkg/providers/ovirt/client/ovirt-client.go
machacekondra: Right, thanks. I will fix.

pkg/providers/ovirt/client/ovirt-client.go
machacekondra: Right, thanks. I will fix. I'll get rid of all 'Must' methods, I used them for testing only, and forgot to remove them all.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Yes, makes sense, thank you.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Yes, makes sense, thank you.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -117,25 +128,63 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
 		return reconcile.Result{}, err
 	}
 
+	if _, ok := ovirtSecretDataMap["caCert"]; !ok {
+		return reconcile.Result{}, fmt.Errorf("oVirt secret must contain caCert attribute")
+	}
+	if len(ovirtSecretDataMap["caCert"]) == 0 {
+		return reconcile.Result{}, fmt.Errorf("oVirt secret caCert cannot be empty")
+	}
+	caCert, err := base64.StdEncoding.DecodeString(ovirtSecretDataMap["caCert"])
```
pkliczewski, line 148: I wonder why not to use `data` instead of `stringData`. It would make k8s to decode the content automatically.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -117,25 +128,63 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
 		return reconcile.Result{}, err
 	}
 
-	diskAttachmentsLink, _ := vm.DiskAttachments()
-	diskAttachments, _ := ovirtConnection.FollowLink(diskAttachmentsLink)
-	dvs := createDVmap(diskAttachments.(*ovirtsdk.DiskAttachmentSlice), ovirtSecret)
-	vmSpec := newVMForCR(vm, dvs, diskAttachments.(*ovirtsdk.DiskAttachmentSlice))
+	// Stop VM
+	vmID, _ := vm.Id()
+	err = ovirt.StopVM(vmID)
```
pkliczewski, line 181: We need to memorize original vm status maybe as annotation. We need it to restore status if import operation fails.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -175,26 +255,98 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
-		return nil, err
+	running := false
+	name, _ := vm.Name()
+	if vmImport.Spec.TargetVMName != nil {
+		name = *vmImport.Spec.TargetVMName
 	}
-	vms, _ := response.Vms()
-	if len(vms.Slice()) > 0 {
-		return vms.Slice()[0], nil
+	return &kubevirtv1.VirtualMachine{
```
pkliczewski, line 323: We need to use common template here matched by source operating system or guest operating system. When using OS.Type we need to use pattern matching to understand which template should be used as base for target vm.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -146,22 +195,53 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+			return reconcile.Result{}, err
+		}
+
+		// CM containing CA for the image import:
+		if err = ensureDVConfigmapExists(r.client, instance, vmSpec.Namespace, ovirtSecretDataMap); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		// Import disks:
+		dvs := createDVmap(vm, ovirtSecretDataMap, instance.Namespace)
```
pkliczewski, line 217: It would be good to `SetControllerReference` on each `DataVolume` which would help us to get cancellation logic for free if VM import object was removed. We could unset it when the import is done.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -207,45 +359,34 @@ func (r *ReconcileVirtualMachineImport) fetchOvirtSecret(vmImport *v2vv1alpha1.V
...
-						SecretRef:     "ovirt-key", // FIXME, should be created dynamically
-						CertConfigMap: "ovirt-ca",  // FIXME, should be created dynamically
+						URL:           creds["apiUrl"],
+						DiskID:        diskID,
+						SecretRef:     ovirtSecret,
+						CertConfigMap: ovirtConfigmap,
 					},
 				},
-				// TODO: Would be great to add storageClassName which we should get from the mapping.
+				// TODO: Should be done according to mappings
```
pkliczewski, line 389: What we need to get here is `storageClass`.

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -2,13 +2,23 @@ package ovirtclient
...
 import (
 	"fmt"
+	"time"
 
 	ovirtsdk "github.com/ovirt/go-ovirt"
 )
 
+const (
+	// Number of minutes to wait for VM to be stopped
+	vmStopTimeout = 5
```
pkliczewski, line 12: Can you add TODO to make them configurable? Maybe config? This can be done when needed not now.

jakub-dzon (approved):

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -100,19 +170,38 @@ func (client *richOvirtClient) fetchVM(id *string, name *string, cluster *string
...
+		// If user specified clusterID, iterate over list of VMs and find the VM
+		// that match the clusterID
+		if clusterID != nil {
+			for _, vm := range vms.Slice() {
+				if vmID, _ := vm.Id(); vmID == *clusterID {
+					return vm, nil
+				}
+			}
+		}
+		return nil, fmt.Errorf("Found more than one virtual machine with name %v in cluster %v(%v). VM IDs: %v", *name, clusterName, clusterID, getVMIDs(vms.Slice()))
```
pkliczewski, line 202: In [formatting](https://golang.org/pkg/fmt/) we can use %s for strings.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: I think this can be done in separate PR.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -175,26 +255,98 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+	if err != nil && errors.IsNotFound(err) {
+		dvSecret := createDVSecret(ovirtSecretDataMap, instance)
+		return client.Create(context.TODO(), dvSecret)
+	}
+	return err
+}
+
+func ensureDVConfigmapExists(client client.Client, instance *v2vv1alpha1.VirtualMachineImport, namespace string, ovirtSecretDataMap map[string]string) error {
+	cm := &corev1.ConfigMap{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: ovirtConfigmap, Namespace: namespace}, cm)
```
masayag, line 270: Shouldn't this name  be unique per ovirt provider instance and not same name for all ovirt providers since its content is provider-specific ?

machacekondra: Not sure if per ovirt provider. Should we also take into account direct execution of the API not using UI? Definitelly we have to create different cm/secret for different credentials, so here using name as constant is wrong.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
masayag: There are 2 issues here:
1. How to support multiple providers within the same key: data expects the input to be base64 encoded, it will not support providing a key, the same way as stringData does. we'll have to define two sets of variables on the same level prefixed differently for ovirt_ and vmware_ which isn't aligned with the design (differentiate between providers by sectiions)

2. How to represent multi-line value in yaml - seems the convention is to represent it as a wrapped base64 string which is the reason for doing it.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
masayag: In the discussion over the document we concluded the vm status shouldn't be changed on failure at provider's end.
Meaning we shouldn't start it if the process failed. I'll verify this point.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
masayag: we'd like also to add a label to all of the resources created by the process (vm/dvs) :
`vmimport.v2v.kubevirt.io/tracker: ...` - but that can be added later on, based on the existence of the label.
The idea is that if that label was set for the vmimport cr, we should propagate it to the resources that were created by as a result (dvs secret/configmap seem less important in that context)

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
masayag: I meant we can use same one for a single ovirt provider. But the name should vary for other ovirt providers.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: what is wrong with using `|`? It seems to be used quite often in Ansible context. Requiring some values encoded and some not seems confusing.

masayag (approved):

PR #18, opened 2020-03-23
Network mapping validator

jakub-dzon:
This PR introduces network mapping rules completeness validation.

### DIFF
--- a/pkg/providers/ovirt/validation/validators/network-mapping-validator.go
+++ b/pkg/providers/ovirt/validation/validators/network-mapping-validator.go
@@ -30,18 +30,19 @@ func NewNetworkMappingValidator(kubevirt kubecli.KubevirtClient) NetworkMappingV
 }
 
 //ValidateNetworkMapping validates network mapping
-func (v *NetworkMappingValidator) ValidateNetworkMapping(nics []*ovirtsdk.Nic, mapping *[]v2vv1alpha1.ResourceMappingItem, defaultNamespace string) []ValidationFailure {
+func (v *NetworkMappingValidator) ValidateNetworkMapping(nics []*ovirtsdk.Nic, mapping *[]v2vv1alpha1.ResourceMappingItem, crNamespace string) []ValidationFailure {
 	var failures []ValidationFailure
 	// Check whether mapping for network is required and was provided
-	if mapping == nil && len(nics) > 0 {
-		failures = append(failures, ValidationFailure{
-			ID:      NetworkMappingID,
-			Message: "Network mapping is missing",
-		})
+	if mapping == nil {
+		if v.hasAtLeastOneWithVNicProfile(nics) {
+			failures = append(failures, ValidationFailure{
+				ID:      NetworkMappingID,
+				Message: "Network mapping is missing",
+			})
+		}
 		return failures
 	}
-	// Get all networks needed by the VM as slice of sources
-	requiredNetworks := getRequiredNetworks(nics)
+
 	mapByID := make(map[string]v2vv1alpha1.ResourceMappingItem)
 	mapByName := make(map[string]v2vv1alpha1.ResourceMappingItem)
 	// Map source id and name to ResourceMappingItem
@@ -53,6 +54,9 @@ func (v *NetworkMappingValidator) ValidateNetworkMapping(nics []*ovirtsdk.Nic, m
 			mapByName[*item.Source.Name] = item
 		}
 	}
+
+	// Get all networks needed by the VM as slice of sources
+	requiredNetworks := getRequiredNetworks(nics)
 	requiredTargetsSet := make(map[v2vv1alpha1.ObjectIdentifier]*string)
 	// Validate that all vm networks are mapped and populate requiredTargetsSet for target existence check
 	for _, network := range requiredNetworks {
@@ -81,19 +85,28 @@ func (v *NetworkMappingValidator) ValidateNetworkMapping(nics []*ovirtsdk.Nic, m
 		if networkType == nil {
 			continue
 		}
-		if failure, valid := v.validateNetwork(networkID, *networkType, defaultNamespace); !valid {
+		if failure, valid := v.validateNetwork(networkID, *networkType, crNamespace); !valid {
 			failures = append(failures, failure)
 		}
 	}
 	return failures
 }
 
-func (v *NetworkMappingValidator) validateNetwork(networkID v2vv1alpha1.ObjectIdentifier, networkType string, defaultNamespace string) (ValidationFailure, bool) {
+func (v *NetworkMappingValidator) hasAtLeastOneWithVNicProfile(nics []*ovirtsdk.Nic) bool {
+	for _, nic := range nics {
+		if _, ok := nic.VnicProfile(); ok {
+			return true
+		}
+	}
+	return false
+}
+
+func (v *NetworkMappingValidator) validateNetwork(networkID v2vv1alpha1.ObjectIdentifier, networkType string, crNamespace string) (ValidationFailure, bool) {
 	switch networkType {
 	case "pod":
 		return ValidationFailure{}, true
 	case "multus":
-		return v.isValidMultusNetwork(networkID, defaultNamespace)
+		return v.isValidMultusNetwork(networkID, crNamespace)
 	default:
 		return ValidationFailure{
 			ID:      NetworkTypeID,
@@ -102,8 +115,8 @@ func (v *NetworkMappingValidator) validateNetwork(networkID v2vv1alpha1.ObjectId
 	}
 }
 
-func (v *NetworkMappingValidator) isValidMultusNetwork(networkID v2vv1alpha1.ObjectIdentifier, defaultNamespace string) (ValidationFailure, bool) {
-	namespace := defaultNamespace
+func (v *NetworkMappingValidator) isValidMultusNetwork(networkID v2vv1alpha1.ObjectIdentifier, crNamespace string) (ValidationFailure, bool) {
+	namespace := crNamespace
 	if networkID.Namespace != nil {
 		namespace = *networkID.Namespace
 	}
@@ -111,7 +124,7 @@ func (v *NetworkMappingValidator) isValidMultusNetwork(networkID v2vv1alpha1.Obj
 	if err != nil {
 		return ValidationFailure{
 			ID:      NetworkTargetID,
-			Message: fmt.Sprintf("Network %s/%s has not been found. Error: %v", namespace, networkID.Name, err),
+			Message: fmt.Sprintf("Network Attachment Defintion %s has not been found. Error: %v", ToLoggableResourceName(networkID.Name, &namespace), err),
 		}, false
 	}
 
@@ -121,14 +134,9 @@ func (v *NetworkMappingValidator) isValidMultusNetwork(networkID v2vv1alpha1.Obj
 func getRequiredNetworks(nics []*ovirtsdk.Nic) []v2vv1alpha1.Source {
 	sourcesSet := make(map[v2vv1alpha1.Source]bool)
 	for _, nic := range nics {
-		if network, ok := nic.Network(); ok {
-			if src, ok := maybeCreateSource(network); ok {
-				sourcesSet[*src] = true
-			}
-		}
 		if vnic, ok := nic.VnicProfile(); ok {
 			if network, ok := vnic.Network(); ok {
-				if src, ok := maybeCreateSource(network); ok {
+				if src, ok := createSourceNetworkIdentifier(network); ok {
 					sourcesSet[*src] = true
 				}
 			}
@@ -141,7 +149,7 @@ func getRequiredNetworks(nics []*ovirtsdk.Nic) []v2vv1alpha1.Source {
 	return sources
 }
 
-func maybeCreateSource(network *ovirtsdk.Network) (*v2vv1alpha1.Source, bool) {
+func createSourceNetworkIdentifier(network *ovirtsdk.Network) (*v2vv1alpha1.Source, bool) {
 	id, okID := network.Id()
 	name, okName := network.Name()
 	if okID || okName {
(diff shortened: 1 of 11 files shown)


pkg/providers/ovirt/validation/validators/network-mapping-validator.go
```diff
@@ -0,0 +1,154 @@
...
+func (v *NetworkMappingValidator) isValidMultusNetwork(networkID v2vv1alpha1.ObjectIdentifier, defaultNamespace string) (ValidationFailure, bool) {
+	namespace := defaultNamespace
+	if networkID.Namespace != nil {
+		namespace = *networkID.Namespace
+	}
+	_, err := v.provider.Find(networkID.Name, namespace)
+	if err != nil {
+		return ValidationFailure{
+			ID:      NetworkTargetID,
+			Message: fmt.Sprintf("Network %s/%s has not been found. Error: %v", networkID.Name, namespace, err),
```
pkliczewski: It is commonly used to have namespace/name.

pkg/providers/ovirt/validation/vm-import-validation.go
```diff
@@ -106,27 +111,68 @@ type VirtualMachineImportValidator struct {
...
 	return VirtualMachineImportValidator{
-		Validator: &validators.ValidatorWrapper{},
+		Validator: validators.NewValidatorWrapper(kubevirtClient),
 		client:    client,
 	}
 }
 
 // Validate validates whether VM described in VirtualMachineImport can be imported
-func (validator *VirtualMachineImportValidator) Validate(vm *ovirtsdk.Vm, vmiCrName *types.NamespacedName) error {
+func (validator *VirtualMachineImportValidator) Validate(vm *ovirtsdk.Vm, vmiCrName *types.NamespacedName, mappings *v2vv1alpha1.OvirtMappings) error {
```
pkliczewski, line 122: I do not see a place where we merge partial mapping from VM import object with global `OvirtMappings`. VM import specific can override global mapping we should merge them before validating.

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
jakub-dzon: Good catch!

pkg/providers/ovirt/validation/vm-import-validation.go
jakub-dzon: That's right - the code hasn't been plugged into reconciliation loop yet. The mapping will be merged there, before being passed to validation.

pkg/providers/ovirt/validation/validators/network-mapping-validator_test.go
```diff
@@ -0,0 +1,235 @@
...
+	podType          = "pod"
+	multusType       = "multus"
+	networkName      = "some-net"
+	networkID        = "some-net-id"
+	wrongNetworkID   = "some-net-bad-id"
+	wrongNetworkName = "some-net-bad"
+
+	targetNetworkName      = "targetNetwork"
+	targetNetworkNamespace = "targetNamespace"
+	findNadMock            func() (*netv1.NetworkAttachmentDefinition, error)
```
masayag: perhaps use short names as a replacement of the full name instead of nad, i.e. NetAttachDef  (net-attach-def is the shorter version of the name in its crd) - then the functons and variables will have more meaningful name.
findNetAttachDefMock, mockNetAttachDefProvider and so on

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
```diff
@@ -0,0 +1,154 @@
...
+		}, false
+	}
+
+	return ValidationFailure{}, true
+}
+
+func getRequiredNetworks(nics []*ovirtsdk.Nic) []v2vv1alpha1.Source {
+	sourcesSet := make(map[v2vv1alpha1.Source]bool)
+	for _, nic := range nics {
+		if network, ok := nic.Network(); ok {
```
masayag: I think that nic.Network() is deprecated and we should rely on vnic profile only to obtain the network name.
Since we populated the vnicprofile in ovirt-client, we should rely on it to provide the details of the VM.

so in this context., I'd remove this block (lines 124-128)

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+	var sources []v2vv1alpha1.Source
+	for source := range sourcesSet {
+		sources = append(sources, source)
+	}
+	return sources
+}
+
+func maybeCreateSource(network *ovirtsdk.Network) (*v2vv1alpha1.Source, bool) {
```
masayag: I think this should be renamed `createSourceNetworkIdentifier` or `createSourceNetworkMapping` without maybe. 
Either we create it, or we return values that indicates the failure to create one.

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
```diff
@@ -0,0 +1,154 @@
...
+		return v.isValidMultusNetwork(networkID, defaultNamespace)
+	default:
+		return ValidationFailure{
+			ID:      NetworkTypeID,
+			Message: fmt.Sprintf("Network %s has unsupported network type: %s", ToLoggableResourceName(networkID.Name, networkID.Namespace), networkType),
+		}, false
+	}
+}
+
+func (v *NetworkMappingValidator) isValidMultusNetwork(networkID v2vv1alpha1.ObjectIdentifier, defaultNamespace string) (ValidationFailure, bool) {
```
masayag: `defaultNamespace` is slightly confusing between the `default` namespace, while it should represent the namespace of the current vmimport CR.

Either we name it simply `namespace` or `crNamespace` which better reflects its role, although less important in the context of this method.

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
```diff
@@ -0,0 +1,154 @@
...
+	namespace := defaultNamespace
+	if networkID.Namespace != nil {
+		namespace = *networkID.Namespace
+	}
+	_, err := v.provider.Find(networkID.Name, namespace)
+	if err != nil {
+		return ValidationFailure{
+			ID:      NetworkTargetID,
+			Message: fmt.Sprintf("Network %s/%s has not been found. Error: %v", namespace, networkID.Name, err),
```
masayag: The message could be more descriptive with a description of the network object, either 'Network Attachment Definition' or 'Target Network' to emphasis this is k8s resource.

masayag: Can ToLoggableResourceName be used here ?

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
```diff
@@ -0,0 +1,154 @@
...
+			Client: kubevirt.NetworkClient(),
+		},
+	}
+}
+
+//ValidateNetworkMapping validates network mapping
+func (v *NetworkMappingValidator) ValidateNetworkMapping(nics []*ovirtsdk.Nic, mapping *[]v2vv1alpha1.ResourceMappingItem, defaultNamespace string) []ValidationFailure {
+	var failures []ValidationFailure
+	// Check whether mapping for network is required and was provided
+	if mapping == nil && len(nics) > 0 {
```
masayag: there is one case of having a single nic not attached to any network which is supported in ovirt.
in that case, we said we'll not create a network device to the vm and we'll add a warning.
perhaps just to be on the safe side, let's make sure that there is at list one nic with vnic-profile set (at this step, it isn't important if the vnic profile is valid or not).

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
```diff
@@ -0,0 +1,154 @@
...
+		failures = append(failures, ValidationFailure{
+			ID:      NetworkMappingID,
+			Message: "Network mapping is missing",
+		})
+		return failures
+	}
+	// Get all networks needed by the VM as slice of sources
+	requiredNetworks := getRequiredNetworks(nics)
```
masayag: we can push that after the creation of  the mapBy* to line 56, where there is going to be a use of it.
Unless there is a reason why it should be executed here.

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
```diff
@@ -0,0 +1,154 @@
...
+				continue
+			}
+		}
+		failures = append(failures, ValidationFailure{
+			ID:      NetworkMappingID,
+			Message: fmt.Sprintf("Required source network '%s' lacks mapping", ToLoggableID(network.ID, network.Name)),
+		})
+	}
+
+	// Validate that all target networks needed by the VM exist in k8s
```
masayag, line 83: the rich client took care to update the full vm entity with its nics, and vnic profile for each nic and the network for each vnic profile.

so the vm spec creation could simply iterate over the vm nics and find the matching target network while the source name/id are available without a need to figure them again, right ?

pkg/providers/ovirt/validation/vm-import-validation.go
masayag: exactly. the input for this method will be the merged mappings where the embedded mappings overrides the master mappings.

masayag (changes requested):

pkg/providers/ovirt/validation/validators/network-mapping-validator_test.go
jakub-dzon: :+1:

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
jakub-dzon: Thanks! That will be s great simplification.

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
jakub-dzon: Will do.

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
jakub-dzon: :+1:

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
jakub-dzon: Yes. It wasn't used because the fail safe logic of that function was not needed. I will use it for consistency.

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
jakub-dzon: :+1:

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
jakub-dzon: Yes, that's correct.

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
jakub-dzon: I've pushed the change, please have a look if that's what you had in mind.

masayag (approved):
Great patch! Thanks.

jakub-dzon:
Thank you!

pkliczewski (approved):

pkliczewski:
@machacekondra What do you think?

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
```diff
@@ -0,0 +1,164 @@
...
+		},
+	}
+}
+
+//ValidateNetworkMapping validates network mapping
+func (v *NetworkMappingValidator) ValidateNetworkMapping(nics []*ovirtsdk.Nic, mapping *[]v2vv1alpha1.ResourceMappingItem, crNamespace string) []ValidationFailure {
+	var failures []ValidationFailure
+	// Check whether mapping for network is required and was provided
+	if mapping == nil {
+		if v.hasAtLeastOneWithVNicProfile(nics) {
```
machacekondra, line 37: Well, so we don't care about nic, which has `empty` VNICs, or do we want to create also rules for `empty` networks. Actually I don't really know why `empty` networks was created in oVirt and what's purpose of them, so not sure if it has any value really. So any reason to fail only in case vnicprofile don't exist?

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
```diff
@@ -0,0 +1,164 @@
...
+			if _, ok := nics[0].VnicProfile(); ok {
```
machacekondra: why you take here `nics[0]`?

jakub-dzon: Good catch! It's a mistake - that check shouldn't be there at all - hasAtLeastOneWithVNicProfile does what I thought was needed.

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
jakub-dzon: Actually, my understanding was that in case described by Moti the nic that is unplugged wouldn't have vnic_profile. Are you saying, that it would have the profile, but the network there wouldn't be present and the check should be `hasAtLeastOneWithNetworkInVNicProfile`?
I wanted to:
 - fail if there is at least one nic that is connected to a network and we have `nil` mapping;
 - allow for configuration with no nics connected to networks and `nil` mapping.

machacekondra (approved):

pkg/providers/ovirt/validation/validators/network-mapping-validator.go
masayag: empty network / empty vnic profile should not be represented in the target VM.
Simply since the API doesn't support that: The network is mandatory to exist:
https://kubevirt.io/api-reference/v0.26.2/definitions.html#_v1_interface
We should add a warning for this case  - as we discussed originally when reviewed the mapping rules.

Regarding why empty networks was created in oVirt:
The vNIC has two properties: plug and link-state.
plug can be on/off i.e. plugged/unplugged.
link-state can be up/down. So if it is down, the vm remains configured as connected to the network, but temporarily the vm is disconnected from it, without the need to remove the definition of the vnic interface and loose the assigned mac address.

PR #21, opened 2020-03-25
Storage mapping validation

jakub-dzon:
Adds storage domain to storage class mapping validation.

### DIFF
--- a/pkg/providers/ovirt/validation/validators/storage-class-provider.go
+++ b/pkg/providers/ovirt/validation/validators/storage-class-provider.go
@@ -1,0 +1,17 @@
+package validators
+
+import (
+	v1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	storage "k8s.io/client-go/kubernetes/typed/storage/v1"
+)
+
+// StorageClasses is responsible for finding storage classes
+type StorageClasses struct {
+	Client storage.StorageClassInterface
+}
+
+// Find retrieves storage class with provided name
+func (finder *StorageClasses) Find(name string) (*v1.StorageClass, error) {
+	return finder.Client.Get(name, metav1.GetOptions{})
+}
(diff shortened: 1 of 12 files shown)


pkg/providers/ovirt/validation/vm-import-validation.go
```diff
@@ -159,22 +160,27 @@ func (validator *VirtualMachineImportValidator) processMappingValidationFailures
...
-		return fmt.Errorf("Mapping rules validation failed for %v. Reasons: %s", vmiCrName, message)
+	} else {
+		updateCondition(&copy.Status.Conditions, validationCompletedReason, "Validating completed successfully", true, v2vv1alpha1.Validating)
 	}
 
+	err = validator.client.Status().Update(context.TODO(), copy)
+	if err != nil {
+		return err
+	}
+	if invalid {
```
pkliczewski: Why not to check `invalid` one time and perform `updateCondition` and status update in specific if branch.

jakub-dzon: That would duplicate the code responsible for the update call and its error handling in both branches.

pkliczewski: What about extracting it to a method?

jakub-dzon: Having the update call in both branches will result in exactly the same number of lines as it is now:) The extracted method will only add to it.
I must admit, that when I look at the code changed as you suggested in the first comment, it does look more elegant, despite the duplication. I will change it this way.

machacekondra (approved):

pkliczewski (approved):

pkg/providers/ovirt/validation/validators/storage-class-provider.go
```diff
@@ -0,0 +1,17 @@
+package validators
+
+import (
+	v1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	storage "k8s.io/client-go/kubernetes/typed/storage/v1"
+)
+
+// StorageClasses is responsible for finding storage classes
+type StorageClasses struct {
```
masayag, line 10: if that the only responsibility of it, perhaps `StorageClassFinder`
if this is expected to grow, no need to rename

jakub-dzon: I don't think it will grow in the future - I will rename it.

PR #22, opened 2020-03-25
Use validation in the reconcile loop

jakub-dzon:
This PR makes use of the validation logic in the reconciliation loop.
The resource mapping preparation will be implemented in a separate PR.

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -122,7 +122,7 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 		return reconcile.Result{}, err
 	}
 
-	provider, err := createProvider(instance.Spec.Source)
+	provider, err := r.createProvider(instance)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
@@ -137,6 +137,26 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 	err = provider.LoadVM(instance.Spec.Source)
 	if err != nil {
 		return reconcile.Result{}, err
+	}
+
+	// Load the external resource mapping
+	resourceMapping, err := r.fetchResourceMapping(&instance.Spec.ResourceMapping)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Prepare/merge the resourceMapping
+	err = provider.PrepareResourceMapping(resourceMapping, instance.Spec.Source)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Validate if it's needed at this stage of processing
+	if shouldValidate(&instance.Status) {
+		err = provider.Validate()
+		if err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
 	// Stop VM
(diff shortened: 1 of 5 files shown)


pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -170,7 +175,13 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-
+	mappings, err := prepareResourceMapping(&instance.Spec)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if err := r.validator.Validate(vm, &request.NamespacedName, mappings); err != nil {
```
pkliczewski: How do you make sure we validate mappings once and not every cycle of the loop?

jakub-dzon: I don't; the input can change any time - the vm, the mapping resource and the import CR. In the most positive scenario we trigger validation once and it passes. With validation failures we actually want to re-validate in every subsequent iteration.

pkliczewski: how do we check whether it was positive scenario?

jakub-dzon: The failure is reported as error returned from the `r.validator.Validate` call; if it's nil, we continue with creating VM, otherwise that iteration is stopped and the reconciliation will be executed again, as  the CR is re-queued if error is returned from the `Reconcile` function.

pkliczewski: I was more thinking about further steps. When we watch dvs or vmi the `Reconcile` is called. How do we make sure that monitoring cycles won't trigger validation again?

jakub-dzon: Oh, OK.
We can potentially base that on the status of the conditions or existence of the vm that is supposed to be created. 

That seems to be a bit wider subject than just validation - what if someone changes the import CR (i.e. network mapping) when the vm exists? We should probably re-validate and update the vm in CNV. If someone changes the storage class, should we do anything with data volume that is currently being transferred to k8s? It looks like o good case for a state machine based on conditions and/or state of sub-processes.

pkliczewski: In general IMS blocks changes to the mappings during migration. Once VM and DVs are created there is no need to validate or react to changes.

jakub-dzon: If we can assume that the mappings mustn't change after the processing starts, checking if `MappingRulesChecking` completed successfully should be sufficient.

jakub-dzon:
I will re-work it after #23 is merged.

jakub-dzon:
The code has a few TODOs that will be implemented in separate PRs.

pkliczewski (approved):
Looks like good stepping stone.

machacekondra (approved):

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -139,6 +139,26 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+
+	// Prepare/merge the resourceMapping
+	err = provider.PrepareResourceMapping(resourceMapping, instance.Spec.Source)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Validate if it's needed at this stage of processing
+	if shouldValidate(&instance.Status) {
+		err = provider.Validate()
```
masayag, line 156: I intended to update the status within this context instead on providers, so we'll have a unified  behavior to all providers.
I'm okay to accept this as a template and re-evaluate later on when we'll start updating the status

PR #23, opened 2020-03-25
Refactor Reconcile method

masayag:
The purpose of the PR is to remove any notation of specific provider
from the controller, excpet when identifying the provider type in
context.

### DIFF
--- a/pkg/providers/ovirt/provider.go
+++ b/pkg/providers/ovirt/provider.go
@@ -1,0 +1,293 @@
+package ovirtprovider
+
+import (
+	"crypto/md5"
+	"encoding/hex"
+	"fmt"
+	"strconv"
+
+	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
+	provider "github.com/kubevirt/vm-import-operator/pkg/providers"
+	ovirtclient "github.com/kubevirt/vm-import-operator/pkg/providers/ovirt/client"
+	ovirtsdk "github.com/ovirt/go-ovirt"
+	yaml "gopkg.in/yaml.v2"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/providers/ovirt/provider.go
```diff
@@ -0,0 +1,293 @@
...
+		KeySecret:     o.ovirtSecretDataMap["password"],
+
+		// TODO: name of the two attributes should be unique per vmimport cr
+		// assuming we wish to GC all of the resources related to this cr
+		ConfigMapName: ovirtSecret + suffix,
+		SecretName:    ovirtConfigmap + suffix,
+	}
+}
+
+func createSuffix(s string) string {
```
machacekondra, line 58: Should we create this suffix by ourselfs, or should we use generate name and label the objects, and later fetch them by labels and remove them?

machacekondra (approved):

pkg/providers/provider.go
```diff
@@ -0,0 +1,30 @@
...
+	corev1 "k8s.io/api/core/v1"
+	kubevirtv1 "kubevirt.io/client-go/api/v1"
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
+)
+
+// Provider defines the methods required by source providers for importing a VM
+type Provider interface {
+	Connect(*corev1.Secret) error
+	Close()
+	SetVM(v2vv1alpha1.VirtualMachineImportSourceSpec) error
```
jakub-dzon: What do you think about calling it more provider-like, i.e. `LoadVM` or `FetchVM`?

jakub-dzon (commented):
Really nice refactoring!

pkg/providers/ovirt/provider.go
masayag: GenerateName seems like the conventional way to approach.
That still requires to provide a value for that label that can be identified by the controller for the CR.

For that we need to agree if we want the configmap/secret created to be per CR, or to let it be GC'ed by the user/client

jakub-dzon (approved):
:+1:

pkliczewski (approved):

PR #25, opened 2020-03-27
Retrieval and merging of resource mappings

jakub-dzon:
This PR adds resource mapping CR retrieval and merging with import CR-provided resource mappings.

### DIFF
--- a/pkg/mappings/resource-mappings.go
+++ b/pkg/mappings/resource-mappings.go
@@ -1,0 +1,31 @@
+package mappings
+
+import (
+	"context"
+
+	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ResourceMappingsFinder provides functionality of retrieving Resource Mapping CRs
+type ResourceMappingsFinder struct {
+	client client.Client
+}
+
+// NewResourceMappingsFinder creates new ResourceMappingsFinder configured with given client
+func NewResourceMappingsFinder(client client.Client) ResourceMappingsFinder {
+	return ResourceMappingsFinder{
+		client: client,
+	}
+}
+
+// GetResourceMapping retrieves current version of a resource mapping CR with given namespaced name
+func (m *ResourceMappingsFinder) GetResourceMapping(namespacedName types.NamespacedName) (*v2vv1alpha1.ResourceMapping, error) {
+	instance := v2vv1alpha1.ResourceMapping{}
+	err := m.client.Get(context.TODO(), namespacedName, &instance)
+	if err != nil {
+		return nil, err
+	}
+	return &instance, nil
+}
--- a/pkg/providers/ovirt/mappings/merger.go
+++ b/pkg/providers/ovirt/mappings/merger.go
@@ -1,0 +1,81 @@
+package mappings
+
+import (
+	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
+	"github.com/kubevirt/vm-import-operator/pkg/utils"
+)
+
+// MergeMappings creates new resource mapping spec containing all of the mappings from the externalMappingSpec with all mappings from crMappings. Mappings from the crMappings will overwrite ones from the external mapping.
+func MergeMappings(externalMappingSpec *v2vv1alpha1.ResourceMappingSpec, vmiMapping *v2vv1alpha1.OvirtMappings) *v2vv1alpha1.OvirtMappings {
+	if externalMappingSpec == nil && vmiMapping == nil {
+		return &v2vv1alpha1.OvirtMappings{}
+	}
+	primaryMappings, secondaryMappings := extractMappings(externalMappingSpec, vmiMapping)
+
+	networkMappings := mergeMappings(primaryMappings.NetworkMappings, secondaryMappings.NetworkMappings)
+	storageMappings := mergeMappings(primaryMappings.StorageMappings, secondaryMappings.StorageMappings)
+
+	ovirtMappings := v2vv1alpha1.OvirtMappings{
+		NetworkMappings: networkMappings,
+		StorageMappings: storageMappings,
+	}
+	return &ovirtMappings
+}
+
+func extractMappings(externalMappingSpec *v2vv1alpha1.ResourceMappingSpec, crMappings *v2vv1alpha1.OvirtMappings) (*v2vv1alpha1.OvirtMappings, *v2vv1alpha1.OvirtMappings) {
+	var primaryMappings, secondaryMappings v2vv1alpha1.OvirtMappings
+	if crMappings != nil {
+		primaryMappings = *crMappings
+	}
+
+	if externalMappingSpec != nil && externalMappingSpec.OvirtMappings != nil {
+		secondaryMappings = *externalMappingSpec.OvirtMappings
+	}
+	return &primaryMappings, &secondaryMappings
+}
+
+func mergeMappings(primaryMappings *[]v2vv1alpha1.ResourceMappingItem, secondaryMappings *[]v2vv1alpha1.ResourceMappingItem) *[]v2vv1alpha1.ResourceMappingItem {
+	var mapping []v2vv1alpha1.ResourceMappingItem
+
+	if primaryMappings == nil {
+		return secondaryMappings
+	}
+	if secondaryMappings == nil {
+		return primaryMappings
+	}
+	secondaryIDMap, secondaryNameMap := utils.IndexByIDAndName(secondaryMappings)
+	usedIDs := make(map[string]bool)
+	// Copy everything from the primary mapping to the output
+	for _, item := range *primaryMappings {
+		id := item.Source.ID
+		name := item.Source.Name
+		if id == nil && name == nil {
+			continue
+		}
+		mapping = append(mapping, item)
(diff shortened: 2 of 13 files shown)


pkg/mappings/resource-mappings.go
```diff
@@ -0,0 +1,31 @@
...
+// NewResourceMappingsFinder creates new ResourceMappingsFinder configured with given client
+func NewResourceMappingsFinder(client client.Client) ResourceMappingsFinder {
+	return ResourceMappingsFinder{
+		client: client,
+	}
+}
+
+// GetResourceMapping retrieves current version of a resource mapping CR with given namespaced name
+func (m *ResourceMappingsFinder) GetResourceMapping(namespacedName types.NamespacedName) (*v2vv1alpha1.ResourceMapping, error) {
+	instance := v2vv1alpha1.ResourceMapping{}
```
pkliczewski, line 25: why not to create a pointer here:

```
instance := &v2vv1alpha1.ResourceMapping{}
```

We would not need to make it a pointer in line 26 and 30.

pkg/mappings/resource-mappings.go
```diff
@@ -0,0 +1,31 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ResourceMappingsFinder provides functionality of retrieving Resource Mapping CRs
+type ResourceMappingsFinder struct {
+	client client.Client
+}
+
+// NewResourceMappingsFinder creates new ResourceMappingsFinder configured with given client
+func NewResourceMappingsFinder(client client.Client) ResourceMappingsFinder {
```
pkliczewski, line 17: It seems like we are not using `NewResourceMappingsFinder`. Was that intended?

pkg/providers/ovirt/mappings/merger.go
```diff
@@ -0,0 +1,83 @@
...
+	} else {
+		primaryMappings = crMappings
+	}
+
+	if externalMappingSpec == nil {
+		secondaryMappings = &v2vv1alpha1.OvirtMappings{}
+	} else {
+		secondaryMappings = externalMappingSpec.OvirtMappings
+	}
+	return
```
pkliczewski: it is unfinished `return`?

pkg/providers/ovirt/mappings/merger.go
```diff
@@ -0,0 +1,83 @@
...
+		name := item.Source.Name
+		if id == nil && name == nil {
+			break
+		}
+		mapping = append(mapping, item)
+		// Delete from the secondary what we've already used
+		if id != nil {
+			usedIDs[*id] = true
+			delete(secondaryIDMap, *id)
+		}
```
pkliczewski, line 60: What about using else if here? If we have both name and id we will attempt to delete it twice.

pkg/providers/ovirt/mappings/merger.go
jakub-dzon: Nope, named return (https://tour.golang.org/basics/7)

pkg/mappings/resource-mappings.go
jakub-dzon: No, good catch - that's a bug.

pkg/providers/ovirt/mappings/merger.go
jakub-dzon: In the other if's body data is removed from the "by-name" secondary mapping index. Will add test that shows the use-case for two separate ifs.

pkg/mappings/resource-mappings.go
jakub-dzon: But having it as it is, we explicitly show that we're passing a reference, not value in line 26 and up the call stack in line 30. It seems to be an advised pattern in Go.

pkliczewski: It would be great if you could point me to this pattern well defined.

I see that value vs pointer semantics for return value seems to be inconsistent between `NewResourceMappingsFinder` and `GetResourceMapping`.

jakub-dzon: Sorry for paid content, but here it is: https://learning.oreilly.com/videos/ultimate-go-programming/9780135261651/9780135261651-UGP2_01_02_03_03. In the transcript starting from "So, there's a guideline here that I want to provide because the ampersand is a very powerful readability operator, and you don't wanna walk away from it"

I agree, there is a return inconsistency between here and the `NewResourceMappingsFinder` and having in mind the advice above we should probably always return a pointer (and show it explicitly on the return line). 

All in all, I think that's something we should discuss and agree on and follow one convention - whichever it is. What is your opinion on returning pointer vs. value?

pkliczewski: I think it would be best to find go guidelines and follow them. Go was always strong about best practices.

pkg/providers/ovirt/mappings/merger.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func extractMappings(externalMappingSpec *v2vv1alpha1.ResourceMappingSpec, crMappings *v2vv1alpha1.OvirtMappings) (primaryMappings *v2vv1alpha1.OvirtMappings, secondaryMappings *v2vv1alpha1.OvirtMappings) {
+	if crMappings == nil {
+		primaryMappings = &v2vv1alpha1.OvirtMappings{}
+	} else {
+		primaryMappings = crMappings
+	}
+
+	if externalMappingSpec == nil {
```
masayag: since `ResourceMappingSpec.OvirtMappings` is optional , I'd suggest to check that it isn't nil so condition will look like:
`if externalMappingSpec == nil || externalMappingSpec.OvirtMappings == nil`

pkg/providers/ovirt/mappings/merger.go
```diff
@@ -0,0 +1,83 @@
...
+
+// MergeMappings creates new resource mapping spec containing all of the mappings from the externalMappingSpec with all mappings from crMappings. Mappings from the crMappings will overwrite ones from the external mapping.
+func MergeMappings(externalMappingSpec *v2vv1alpha1.ResourceMappingSpec, vmiMapping *v2vv1alpha1.OvirtMappings) v2vv1alpha1.OvirtMappings {
+	if externalMappingSpec == nil && vmiMapping == nil {
+		return v2vv1alpha1.OvirtMappings{}
+	}
+	primaryMappings, secondaryMappings := extractMappings(externalMappingSpec, vmiMapping)
+
+	networkMappings := mergeMappings(primaryMappings.NetworkMappings, secondaryMappings.NetworkMappings)
+	storageMappings := mergeMappings(primaryMappings.StorageMappings, secondaryMappings.StorageMappings)
```
masayag, line 16: My only comment regarding this - the merged mappings might contain more entries than what is actually needed for the VM import. However, i think the fee is small, comparing how complex this could look if we'll throw in the source VM.

We need to make sure to consider only the resources which are relevant for the VM Import on the validation (e.g. not fail if a network that isn't required by the VMI doesn't exist)

pkg/providers/ovirt/mappings/merger.go
jakub-dzon: Good catch! Thanks!

pkg/providers/ovirt/mappings/merger.go
jakub-dzon: That's how the mapping validator behaves - I will add test case to show that it works.

pkg/providers/ovirt/mappings/merger.go
```diff
@@ -0,0 +1,83 @@
...
+	if secondaryMappings == nil {
+		return primaryMappings
+	}
+	secondaryIDMap, secondaryNameMap := utils.IndexByIDAndName(secondaryMappings)
+	usedIDs := make(map[string]bool)
+	// Copy everything from the primary mapping to the output
+	for _, item := range *primaryMappings {
+		id := item.Source.ID
+		name := item.Source.Name
+		if id == nil && name == nil {
```
machacekondra, line 52: So here if any item in primary mapping don't have specified `name` or `id`, we skip all other items in mappings?

jakub-dzon: That's definitely a bug, should be `continue`.

pkg/mappings/resource-mappings.go
jakub-dzon: What I've found so far is not really a rule, but a hint what might be proper:
- return value;
- return pointer if the returned struct is "big" (when the heap allocation cost gets lower than copying value);
Source: https://philpearl.github.io/post/bad_go_pointer_returns/

pkg/providers/ovirt/mappings/merger.go
```diff
@@ -0,0 +1,80 @@
...
+	if crMappings != nil {
+		primaryMappings = *crMappings
+	}
+
+	if externalMappingSpec != nil && externalMappingSpec.OvirtMappings != nil {
+		secondaryMappings = *externalMappingSpec.OvirtMappings
+	}
+	return &primaryMappings, &secondaryMappings
+}
+func mergeMappings(primaryMappings *[]v2vv1alpha1.ResourceMappingItem, secondaryMappings *[]v2vv1alpha1.ResourceMappingItem) *[]v2vv1alpha1.ResourceMappingItem {
```
masayag, line 37: Pls add a new-line before this function starts

masayag (approved):

pkg/mappings/resource-mappings.go
pkliczewski: Yes, it makes sense.

pkliczewski (approved):

machacekondra (approved):

PR #26, opened 2020-03-29
Normalize VM name

masayag:
The PR follows the same logic as suggested on UI side:
https://github.com/openshift/console/pull/4726/files#diff-db36e0ad662f32cf57375ac6941c817cR108

### DIFF
--- a/pkg/providers/ovirt/provider.go
+++ b/pkg/providers/ovirt/provider.go
@@ -13,6 +13,7 @@ import (
 	provider "github.com/kubevirt/vm-import-operator/pkg/providers"
 	ovirtclient "github.com/kubevirt/vm-import-operator/pkg/providers/ovirt/client"
 	"github.com/kubevirt/vm-import-operator/pkg/providers/ovirt/validation"
+	"github.com/kubevirt/vm-import-operator/pkg/utils"
 	ovirtsdk "github.com/ovirt/go-ovirt"
 	yaml "gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/providers/ovirt/provider.go
```diff
@@ -173,16 +174,21 @@ func (o *OvirtProvider) CreateVMSpec(vmImport *v2vv1alpha1.VirtualMachineImport)
...
+		Namespace: vmImport.Namespace,
+		Labels:    labels,
 	}
+	if shouldGenerate {
+		// TODO: consider a uniquer value for VM name prefix, perhaps crName
+		objectMeta.GenerateName = "ovirt-"
+	} else {
+		objectMeta.Name = name
+	}
+	// TODO: The VM Spec is idempotent, however, we should add a notation by which we should search for the VM (e.g label/ownerReference)
```
pkliczewski, line 189: Should we check whether the vm exists or are we going to rely on creating and failing?

masayag: The assumption is that on UI flow the UI will be responsible to create a unique name and the user will have to adjust it if it is taken. Then the UI should use the `targetVMName` to provider the name. In this context we should fail if the namespace/name is taken. And that should be the case for any use of `targetVMName`

The other case is if the normalized name produces the need to generate a name. In most cases that won't happen, since `shouldGenerate` can be `true` only if the vm name is empty (cannot happen) or comprised of non-alphanumeric characters (not likely).

Question remains: How should we proceed if a normalized-name isn't unique - should we use the normalized name as a prefix to the generateName and then create the VM ?

I'd prefer to have that check on the reconcile loop, since it will be common to all flows and should rely on the status of the vm import.

pkliczewski: Sounds like good approach to me

machacekondra (approved):

jakub-dzon (approved):

pkliczewski (approved):

pkliczewski:
@masayag please rebase so it can be merged

PR #28, opened 2020-03-30
Condition management refactoring

jakub-dzon:
This PR:
 - introduces conditions package that helps with status condition management;

### DIFF
--- a/pkg/conditions/conditions.go
+++ b/pkg/conditions/conditions.go
@@ -1,0 +1,51 @@
+package conditions
+
+import (
+	"time"
+
+	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NewCondition creates condition of a given type in the conditions list with given reason, message and status
+func NewCondition(conditionType v2vv1alpha1.VirtualMachineImportConditionType, reason string, message string, status v1.ConditionStatus) v2vv1alpha1.VirtualMachineImportCondition {
+	now := metav1.NewTime(time.Now())
+	condition := v2vv1alpha1.VirtualMachineImportCondition{
+		Type:               conditionType,
+		LastTransitionTime: &now,
+		LastHeartbeatTime:  &now,
+		Message:            &message,
+		Reason:             &reason,
+		Status:             status,
+	}
+	return condition
+}
+
+// UpsertCondition updates or creates condition in the virtualMachineImportStatus
+func UpsertCondition(vmi *v2vv1alpha1.VirtualMachineImport, condition v2vv1alpha1.VirtualMachineImportCondition) {
+	existingCondition := FindConditionOfType(vmi.Status.Conditions, condition.Type)
+	now := metav1.NewTime(time.Now())
+
+	if existingCondition != nil {
+		existingCondition.Message = condition.Message
+		existingCondition.Reason = condition.Reason
+		existingCondition.LastHeartbeatTime = &now
+		if existingCondition.Status != condition.Status {
+			existingCondition.Status = condition.Status
+			existingCondition.LastTransitionTime = condition.LastTransitionTime
+		}
+	} else {
+		vmi.Status.Conditions = append(vmi.Status.Conditions, condition)
+	}
+}
+
+// FindConditionOfType finds condition of a conditionType type in the conditions slice
+func FindConditionOfType(conditions []v2vv1alpha1.VirtualMachineImportCondition, conditionType v2vv1alpha1.VirtualMachineImportConditionType) *v2vv1alpha1.VirtualMachineImportCondition {
+	for i := range conditions {
+		if conditions[i].Type == conditionType {
+			return &conditions[i]
+		}
+	}
+	return nil
+}
--- a/pkg/conditions/conditions_test.go
+++ b/pkg/conditions/conditions_test.go
@@ -1,0 +1,136 @@
+package conditions_test
+
+import (
+	"time"
+
+	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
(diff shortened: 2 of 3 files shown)


pkg/conditions/conditions.go
```diff
@@ -0,0 +1,51 @@
...
+		LastHeartbeatTime:  &now,
+		Message:            &message,
+		Reason:             &reason,
+		Status:             status,
+	}
+	return condition
+}
+
+// UpsertCondition updates or creates condition in the virtualMachineImportStatus
+func UpsertCondition(vmi *v2vv1alpha1.VirtualMachineImport, condition v2vv1alpha1.VirtualMachineImportCondition) {
```
machacekondra: Upsert is a typo I guess?

pkg/conditions/conditions_test.go
```diff
@@ -0,0 +1,135 @@
+package conditions
```
masayag: why not to use `conditions_test` and by that to make sure the code is tested like any consumer from external package that makes no assumption about the tested code and uses only the exported interfaces ?

We haven't agreed on such behavior, but I used that for the utils package and would like to use it for other tests as well.

Another advantage, by looking at the tests, one can get a clear look of how that code should be imported and used.

pkg/conditions/conditions.go
```diff
@@ -0,0 +1,51 @@
...
+			existingCondition.Status = condition.Status
+			existingCondition.LastTransitionTime = condition.LastTransitionTime
+		}
+	} else {
+		vmi.Status.Conditions = append(vmi.Status.Conditions, condition)
+	}
+}
+
+// FindConditionOfType finds condition of a conditionType type in the conditions slice
+func FindConditionOfType(conditionType v2vv1alpha1.VirtualMachineImportConditionType, conditions []v2vv1alpha1.VirtualMachineImportCondition) *v2vv1alpha1.VirtualMachineImportCondition {
```
masayag: I'm not sure if there is a guideline for that anywhere, but I think having the collection to search in first and the search criteria second 'feels' more natural, e.g. append()

pkg/conditions/conditions_test.go
```diff
@@ -0,0 +1,135 @@
...
+		Expect(foundProcessing.Status).To(Equal(status))
+		Expect(foundProcessing.LastHeartbeatTime.Time).To(BeTemporally("<=", time.Now()))
+		Expect(foundProcessing.LastTransitionTime.Time).To(BeTemporally("<=", time.Now()))
+
+	})
+	It("should update condition", func() {
+		oldMessage := "old-message"
+		oldReason := "old-reason"
+		minuteAgo := metav1.NewTime(time.Now().Add(-time.Minute))
+		validating := v2vv1alpha1.VirtualMachineImportCondition{
```
masayag: perhaps rename the `validating` variable to `beforeUpdate` so the expectation checks will be clearer in terms of  which side we're comparing

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -303,9 +315,15 @@ func createDVSecret(creds provider.DataVolumeCredentials, vmImport *v2vv1alpha1.
 	}
 }
 
-func shouldValidate(vmi *v2vv1alpha1.VirtualMachineImportStatus) bool {
-	// TODO: check the status - status manipulation package is needed
-	return true
+func shouldValidate(vmiStatus *v2vv1alpha1.VirtualMachineImportStatus) bool {
+	validingCondition := conditions.FindConditionOfType(v2vv1alpha1.Validating, vmiStatus.Conditions)
```
masayag: Should `validingCondition`  be `validatingCondition` or it is just an abbreviation of it ?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -318,3 +336,37 @@ func (r *ReconcileVirtualMachineImport) createProvider(vmi *v2vv1alpha1.VirtualM
 
 	return nil, fmt.Errorf("Invalid source type. only Ovirt type is supported")
 }
+
+func (r *ReconcileVirtualMachineImport) upsertStatusConditions(vmiName types.NamespacedName, newConditions []v2vv1alpha1.VirtualMachineImportCondition) error {
+	var instance v2vv1alpha1.VirtualMachineImport
+	err := r.client.Get(context.TODO(), vmiName, &instance)
```
masayag: @pkliczewski Isn't there an API call to fetch only the Status of the resource instead of the entire element or did we do some custom code for that in java-kubevirt-client ?

pkliczewski: There is only [StatusClient](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/interfaces.go#L86) interface but it to not provide `Get` nor `List`

pkg/conditions/conditions_test.go
pkliczewski: [Here](https://golang.org/pkg/testing/) we have the answer but we seems not to be consistent with it.

pkliczewski (commented):
You specified 2 bullets in the PR description this feels like 2 PRs. It would be better to understand what changes are responsible for condition management and what is using it.

pkg/conditions/conditions_test.go
masayag: We can discuss the benefit of it vs testing best practices like this [one](https://medium.com/@matryer/5-simple-tips-and-tricks-for-writing-unit-tests-in-golang-619653f90742)

jakub-dzon: I like the idea of using `*_test` packages - the benefits are big for a package user.

pkg/conditions/conditions.go
jakub-dzon: The idea was to transcribe: "find condition of type conditionType in slice conditions". Following similar pattern as elsewhere seems like a good idea, though.

masayag (approved):

pkliczewski (approved):

PR #29, opened 2020-03-30
Add mapper module

machacekondra:
This PR add new mapper module, which contains functionality to map
external VM to kubevirt VM.

This is just refactoring code. No new functionality. Functionality will go in another PR(s).

### DIFF
--- a/pkg/providers/ovirt/mapper/mapper.go
+++ b/pkg/providers/ovirt/mapper/mapper.go
@@ -1,0 +1,210 @@
+package mapper
+
+import (
+	"strconv"
+
+	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
+	provider "github.com/kubevirt/vm-import-operator/pkg/providers"
+	"github.com/kubevirt/vm-import-operator/pkg/utils"
+	ovirtsdk "github.com/ovirt/go-ovirt"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubevirtv1 "kubevirt.io/client-go/api/v1"
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
+)
+
+const (
+	cdiAPIVersion = "cdi.kubevirt.io/v1alpha1"
+)
+
+var biosMapping = map[string]*kubevirtv1.Bootloader{
+	"q35_sea_bios":    &kubevirtv1.Bootloader{BIOS: &kubevirtv1.BIOS{}},
+	"q35_secure_boot": &kubevirtv1.Bootloader{BIOS: &kubevirtv1.BIOS{}},
+	"q35_ovmf":        &kubevirtv1.Bootloader{EFI: &kubevirtv1.EFI{}},
+}
+
+var archMapping = map[string]string{
+	"x86_64": "q35",
+}
+
+// OvirtMapper is struct that holds attributes needed to map oVirt VM to kubevirt VM
+type OvirtMapper struct {
+	vm        *ovirtsdk.Vm
+	mappings  *v2vv1alpha1.OvirtMappings
+	creds     provider.DataVolumeCredentials
+	namespace string
+}
+
+// NewOvirtMapper create ovirt mapper object
+func NewOvirtMapper(vm *ovirtsdk.Vm, mappings *v2vv1alpha1.OvirtMappings, creds provider.DataVolumeCredentials, namespace string) *OvirtMapper {
+	return &OvirtMapper{
+		vm:        vm,
+		mappings:  mappings,
+		creds:     creds,
+		namespace: namespace,
+	}
+}
+
+// MapVM map oVirt API VM definition to kubevirt VM definition
+func (o *OvirtMapper) MapVM(targetVMName *string) *kubevirtv1.VirtualMachine {
+	vmSpec := kubevirtv1.VirtualMachine{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: o.namespace,
+		},
+		Spec: kubevirtv1.VirtualMachineSpec{
+			Template: &kubevirtv1.VirtualMachineInstanceTemplateSpec{
+				Spec: kubevirtv1.VirtualMachineInstanceSpec{
+					Domain: kubevirtv1.DomainSpec{},
+				},
+			},
+		},
+	}
+
+	// Map name
+	vmName, shouldGenerate := o.resolveVMName(targetVMName, o.vm)
+	if shouldGenerate {
+		vmSpec.ObjectMeta.GenerateName = "ovirt-"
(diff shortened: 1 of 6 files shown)


pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -0,0 +1,180 @@
...
+	"q35_sea_bios":    &kubevirtv1.Bootloader{BIOS: &kubevirtv1.BIOS{}},
+	"q35_secure_boot": &kubevirtv1.Bootloader{BIOS: &kubevirtv1.BIOS{}},
+	"q35_ovmf":        &kubevirtv1.Bootloader{EFI: &kubevirtv1.EFI{}},
+}
+
+var archMapping = map[string]string{
+	"x86_64": "q35",
+}
+
+// OvirtMapper is struct that hold attributes needed to map oVirt VM to kubevirt VM
```
masayag: hold --> holds

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -0,0 +1,180 @@
...
+			Template: &kubevirtv1.VirtualMachineInstanceTemplateSpec{
+				Spec: kubevirtv1.VirtualMachineInstanceSpec{
+					Domain: kubevirtv1.DomainSpec{},
+				},
+			},
+		},
+	}
+
+	// Map name
+	vmSpec.ObjectMeta.Name, _ = o.vm.Name()
```
masayag: I think you're missing a rebase on a commit that introduced this code:

https://github.com/kubevirt/vm-import-operator/blob/f7b86e69eefdfc420106db4de3cc1b19bd87e691/pkg/providers/ovirt/provider.go#L178

masayag (changes requested):

pkg/providers/ovirt/mapper/mapper.go
machacekondra: Right, thanks, added

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -0,0 +1,200 @@
...
+			Template: &kubevirtv1.VirtualMachineInstanceTemplateSpec{
+				Spec: kubevirtv1.VirtualMachineInstanceSpec{
+					Domain: kubevirtv1.DomainSpec{},
+				},
+			},
+		},
+	}
+
+	// Map name
+	vmSpec.ObjectMeta.Name, _ = o.resolveVMName(targetVMName, o.vm)
```
masayag: Pls check how it was used originally - https://github.com/kubevirt/vm-import-operator/pull/29/files#diff-92ced7ca268bcb302a8cbee9da107764L178

It returns two values: name and a bool to notate if the name should be generated or use the returned name.
Based on the return value,. vmSpec.ObjectMeta.Name or vmSpec.ObjectMeta.GenerateName are being updated.

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		if mapping.Source.ID != nil {
+			if id, _ := sd.Id(); id == *mapping.Source.ID {
+				return mapping.Target.Name
+			}
+		}
+	}
+
+	return "default" // TODO: Should we use here default, or fail?
```
masayag: iirc, `default` is just an annotation set for the default storageClass and not an actual name.
If storageClass is not specified, we can should rely on the default storageClass (not sure if it should be provided explicitly or DV can be created without stating it).

By not requiring the field, we can simplify the mappings and the input needed by the user.

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -0,0 +1,200 @@
...
+func (o *OvirtMapper) MapDisks() map[string]cdiv1.DataVolume {
+	diskAttachments, _ := o.vm.DiskAttachments()
+	dvs := make(map[string]cdiv1.DataVolume, len(diskAttachments.Slice()))
+
+	for _, diskAttachment := range diskAttachments.Slice() {
+		attachID, _ := diskAttachment.Id()
+		disk, _ := diskAttachment.Disk()
+		diskID, _ := disk.Id()
+		sdClass := o.getStorageClassForDisk(disk, o.mappings)
+		quantity, _ := resource.ParseQuantity(strconv.FormatInt(disk.MustProvisionedSize(), 10))
```
pkliczewski: We could get a panic from `MustProvisionedSize`

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -0,0 +1,200 @@
...
+					Imageio: &cdiv1.DataVolumeSourceImageIO{
+						URL:           o.creds.URL,
+						DiskID:        diskID,
+						SecretRef:     o.creds.SecretName,
+						CertConfigMap: o.creds.ConfigMapName,
+					},
+				},
+				PVC: &corev1.PersistentVolumeClaimSpec{
+					StorageClassName: &sdClass,
+					// TODO: should we have parameter in resourceMapping for this?
```
pkliczewski: Access modes seems to be [dependent](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) on strageclass providers. `ReadWriteOnce` seems to be supported by everyone so I would stick to this mode.

pkg/providers/ovirt/mapper/mapper.go
pkliczewski: Default is not needed. If not provided the `default` is used.

pkg/providers/ovirt/mapper/mapper.go
machacekondra: Done.

pkg/providers/ovirt/mapper/mapper.go
machacekondra: Done. I've removed the storage class from spec, if not target matches.

pkg/providers/ovirt/mapper/mapper.go
machacekondra: Right, changed.

pkg/providers/ovirt/mapper/mapper.go
machacekondra: Right. But maybe user would want to specify the mode. That something how we can improve the mapping API in future.

masayag (approved):

pkliczewski (approved):

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -0,0 +1,210 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubevirtv1 "kubevirt.io/client-go/api/v1"
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
+)
+
+const (
+	cdiAPIVersion = "cdi.kubevirt.io/v1alpha1"
+)
+
+var biosMapping = map[string]*kubevirtv1.Bootloader{
```
jakub-dzon, line 21: It would be good to use this mapping in validation - to have one source of truth.

PR #34, opened 2020-03-31
Map VM attributes

machacekondra:
This PR add mapping to all the core VM attributes.

### DIFF
--- a/pkg/providers/ovirt/mapper/mapper.go
+++ b/pkg/providers/ovirt/mapper/mapper.go
@@ -1,7 +1,9 @@
 package mapper
 
 import (
+	"fmt"
 	"strconv"
+	"strings"
 
 	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
 	provider "github.com/kubevirt/vm-import-operator/pkg/providers"
@@ -15,10 +17,22 @@ import (
 )
 
 const (
-	cdiAPIVersion = "cdi.kubevirt.io/v1alpha1"
+	cdiAPIVersion           = "cdi.kubevirt.io/v1alpha1"
+	customPropertyHugePages = "hugepages"
+	// LabelOrigin define name of the label which holds value of origin attribute of oVirt VM
+	LabelOrigin = "origin"
+	// LabelInstanceType define name of the label which holds value of instance type attribute of oVirt VM
+	LabelInstanceType = "instanceType"
+	// LabelTag define name of the label which holds tags of oVirt VM
+	LabelTag = "tags"
+	// AnnotationComment define name of the annotation which holds comment of oVirt VM
+	AnnotationComment = "comment"
+	// AnnotationSso define name of the annotation which holds sso method of oVirt VM
+	AnnotationSso = "sso"
 )
 
-var biosMapping = map[string]*kubevirtv1.Bootloader{
+// BiosTypeMapping defines mapping of BIOS types between oVirt and kubevirt domains
+var BiosTypeMapping = map[string]*kubevirtv1.Bootloader{
 	"q35_sea_bios":    &kubevirtv1.Bootloader{BIOS: &kubevirtv1.BIOS{}},
 	"q35_secure_boot": &kubevirtv1.Bootloader{BIOS: &kubevirtv1.BIOS{}},
 	"q35_ovmf":        &kubevirtv1.Bootloader{EFI: &kubevirtv1.EFI{}},
@@ -26,6 +40,7 @@ var biosMapping = map[string]*kubevirtv1.Bootloader{
 
 var archMapping = map[string]string{
 	"x86_64": "q35",
+	"ppc64":  "pseries",
 }
 
 // OvirtMapper is struct that holds attributes needed to map oVirt VM to kubevirt VM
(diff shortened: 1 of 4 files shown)


pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -206,5 +279,168 @@ func (o *OvirtMapper) mapCPU() *kubevirtv1.CPU {
...
+		if methods, _ := sso.Methods(); len(methods.Slice()) > 0 {
+			annotations[AnnotationSso] = string(ovirtsdk.SSOMETHOD_GUEST_AGENT)
+		} else {
+			annotations[AnnotationSso] = "disabled"
+		}
+	}
+
+	// Errata
+	var errataValue []string
+	if errata, ok := o.vm.KatelloErrata(); ok {
```
masayag: I don't think we need to map katello errata.
If so, the vm will have to be registered to Satellite/Katello and use it for viewing the content/errata available for the VM.
Also, If we set that as part of the annotation, I'm not aware of a process that will make sure to remove/update when errata is applied for the VM.

pkliczewski (approved):

pkg/providers/ovirt/mapper/mapper.go
machacekondra: Well, it was defined in the document so I did that. I think it's not that important as well, so I've removed it. We can add it later if needed.

pkg/providers/ovirt/validation/validators/vm-validator.go
```diff
@@ -5,10 +5,15 @@ import (
 	"strconv"
 
 	ovirtsdk "github.com/ovirt/go-ovirt"
+	kubevirtv1 "kubevirt.io/client-go/api/v1"
 )
 
 // BiosTypeMapping defines mapping of BIOS types between oVirt and kubevirt domains
-var BiosTypeMapping = map[string]string{"q35_ovmf": "efi", "q35_sea_bios": "bios", "q35_secure_boot": "bios"}
+var BiosTypeMapping = map[string]*kubevirtv1.Bootloader{
```
jakub-dzon: What do you think about keeping it in the mapper and referencing it in the validation?

pkg/providers/ovirt/mapper/mapper_test.go
```diff
@@ -40,6 +44,53 @@ var _ = Describe("Test mapping virtual machine attributes", func() {
 		bootloader := vmSpec.Spec.Template.Spec.Domain.Firmware.Bootloader.BIOS
 		Expect(bootloader).ToNot(BeNil())
 	})
+
+	It("For HA", func() {
```
jakub-dzon: I think that Ginkgo expectation is that user defines what should happen in the `It` function text; i.e. `It("should map HA")` so it sounds naturally as "It should map HA".

jakub-dzon (changes requested):
Nice mapping code! One question and one suggestion.

pkg/providers/ovirt/validation/validators/vm-validator.go
machacekondra: Well, since it's called `BiosTypeMapping` looks like a better package is `mapping` :) so I've moved it.

pkg/providers/ovirt/mapper/mapper_test.go
machacekondra: Yes, that looks better, thanks.

jakub-dzon (approved):
:+1:

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -15,17 +17,32 @@ import (
...
+	cdiAPIVersion           = "cdi.kubevirt.io/v1alpha1"
+	customPropertyHugePages = "hugepages"
+	// LabelOrigin define name of the label which holds value of origin attribute of oVirt VM
+	LabelOrigin = "origin"
+	// LabelInstanceType define name of the label which holds value of instance type attribute of oVirt VM
+	LabelInstanceType = "instanceType"
+	// LabelTag define name of the label which holds tags of oVirt VM
+	LabelTag = "tags"
+	// LabelFQDN define name of the label which holds fqdn of oVirt VM
+	LabelFQDN = "fqdn"
```
masayag, line 29: should we map the `fqdn` to hostname field of  https://kubevirt.io/api-reference/v0.26.4/definitions.html#_v1_virtualmachineinstancespec ?

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -206,5 +283,158 @@ func (o *OvirtMapper) mapCPU() *kubevirtv1.CPU {
...
+		for _, tag := range tags.Slice() {
+			tagName, _ := tag.Name()
+			tagList = append(tagList, tagName)
+		}
+		labels[LabelTag] = strings.Join(tagList, ",")
+	}
+
+	// Fqdn
+	if fqdn, ok := o.vm.Fqdn(); ok {
+		labels[LabelFQDN] = fqdn
```
masayag: should we map the `fqdn` to hostname field of  https://kubevirt.io/api-reference/v0.26.4/definitions.html#_v1_virtualmachineinstancespec ?

machacekondra: Reading the doc of hostname field:

 https://kubevirt.io/user-guide/#/usage/dns?id=creating-unique-dns-entries-per-virtualmachineinstance

it's for internal resolution like <hostname>.<subdomain>. I think we can map it to hostname, so internal resolution will keep working, because if hostname is not specified it fallback to `vmi_name`.

masayag (approved):

PR #35, opened 2020-03-31
Add DiskMappings for specific disk mapping

masayag:
_DiskMappings_ is a method for placement of disks of the same VM on
different storage classes. Using _DiskMappings_ it will be allowed to map
each of source VM disks to storage class.

If a specific mapping for a disk is provided, it will precede the
mapping from _StorageMappings_ if the storage domain as specified in both
mappings lists.

- _DiskMappings_ will fail to map the disk in the following cases:
  - Disk ID/Name doesn't exist for the source VM (ID precedes name)
  - Target storage class or target namespace for the DV doesn't exist
- _DiskMappings_ will be used only if the VM has disk that its identifier
match the value specified in it.
- _DiskMappings.Source.ID_ should hold the ID of the disk and not the ID of
the disk-attachment that describes its association to the VM.
- _DiskMappings_ is optional, if not provided mappings will rely on
_StorageMappings_.
- _DiskMappings_ are supported within VM Import CR only.

### DIFF
--- a/pkg/providers/ovirt/client/ovirt-client.go
+++ b/pkg/providers/ovirt/client/ovirt-client.go
@@ -93,6 +93,14 @@ func (client *richOvirtClient) GetVM(id *string, name *string, cluster *string, 
 	if err != nil {
 		return nil, err
 	}
+	err = client.populateInstanceType(vm)
+	if err != nil {
+		return nil, err
+	}
+	err = client.populateTags(vm)
+	if err != nil {
+		return nil, err
+	}
 	return vm, nil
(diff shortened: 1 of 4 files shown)


pkliczewski:
@suomiy please review

atiratree:
> - DiskMappings.Source.ID should hold the ID of the disk and not the ID of
the disk-attachment that describes its association to the VM. 

maybe this could be mentioned in the types.go ?

otherwise looks good

jakub-dzon (approved):

PR #37, opened 2020-04-01
Ignoring Idea project dir

jakub-dzon:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 # Temporary Build Files
 build/_output
 build/_test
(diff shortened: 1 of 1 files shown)


masayag (approved):
Would it make sense to include also the vendor directory ?

jakub-dzon:
> Would it make sense to include also the vendor directory ?

I feel really bad about having dependencies committed with the actual code, but the reproducible builds upside wins for me: https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory

PR #38, opened 2020-04-01
Update watch events filtering

machacekondra:
This patch update the watch events filtering as follows:

1) For the VMImport CR we ignore update events if import succeded.
2) For the VM resource we ignore all Create/Update events.
3) For the DV resource we ignore all Create/Update events.

We ignore the update events, because import process is one time
operation and we ignore later updates to the CR.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/spf13/pflag"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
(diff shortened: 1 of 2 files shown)


jakub-dzon:
What are the events we are actually interested in for VM and DV?

machacekondra:
> What are the events we are actually interested in for VM and DV?

I think it's delete, because if user delete VM/DV we re-create it.

jakub-dzon:
What is the final criteria for making the import CR successful? Will we do that in response to some event? Disk download takes some time and happens asynchronously to our reconciliation, right?

machacekondra:
Well, there are two I think. First if `startVM` is `true` CR is successfull, if VM is running. The second, if `startVM` is `false`, the successfull state is when DataVolume `phase` is `Succeeded`

machacekondra:
I think we can watch and wait for the succeded state of DV internally, not by CR events, or do you think we shouldn't?

jakub-dzon:
I _feel_ it would be better to rely on cluster mechanisms for scheduling work rather than building our internal probing loop if there is a proper event we can subscribe to. Even if not, we can always re-queue the CR with a desired wait time (`Result.RequeueAfter`) to unplug the standard back-off mechanism. And with that approach we would need to bypass parts of the loop code based on conditions that are recorded within import CR (like validation already).

@masayag, @pkliczewski - what is your opinion?

pkliczewski:
I think that important aspect that we need to pay attention to is parallel imports. There could be many imports in progress at the same time and our controller should be able to handle them. I think we can't block the loop to track single import.

PR #40, opened 2020-04-02
Nic mappings

machacekondra:

### DIFF
--- a/pkg/providers/ovirt/mapper/mapper.go
+++ b/pkg/providers/ovirt/mapper/mapper.go
@@ -122,6 +122,12 @@ func (o *OvirtMapper) MapVM(targetVMName *string) *kubevirtv1.VirtualMachine {
 	// Map timezone
 	vmSpec.Spec.Template.Spec.Domain.Clock = o.mapTimeZone()
 
+	// Map network interfaces
+	vmSpec.Spec.Template.Spec.Domain.Devices.Interfaces = o.mapNics()
+
+	// Map networks
+	vmSpec.Spec.Template.Spec.Networks = o.mapNetworks()
+
 	return &vmSpec
 }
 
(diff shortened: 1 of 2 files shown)


pkliczewski (approved):

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -176,6 +182,75 @@ func (o *OvirtMapper) MapDisks() map[string]cdiv1.DataVolume {
...
+	var kubevirtNics []kubevirtv1.Interface
+	nics, _ := o.vm.Nics()
+	for _, nic := range nics.Slice() {
+		// This network interface don't have any network specified.
+		if _, ok := nic.VnicProfile(); !ok {
+			continue
+		}
+
+		kubevirtNic := kubevirtv1.Interface{}
+		kubevirtNic.Name, _ = nic.Name()
```
jakub-dzon, line 195: Aren't there any restrictions on the name format - like with the vm name? Shouldn't it be normalized?

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -176,6 +182,75 @@ func (o *OvirtMapper) MapDisks() map[string]cdiv1.DataVolume {
...
+	nics, _ := o.vm.Nics()
+	for _, nic := range nics.Slice() {
+		// This network interface don't have any network specified.
+		nicProfile, ok := nic.VnicProfile()
+		if !ok {
+			continue
+		}
+
+		kubevirtNet := o.getNetworkForNic(nicProfile)
+		kubevirtNet.Name, _ = nic.Name()
```
jakub-dzon, line 219: Same question as with the interface, as they have to match.

jakub-dzon:
Also - the build has failed.

pkg/providers/ovirt/mapper/mapper.go
machacekondra: I am not sure, I will check.

machacekondra: oVirt backend already implements the validation and only nic with alphanumeric characters and '-', '_' can be created. So I think we don't need the validation here.

jakub-dzon: :+1:

jakub-dzon (approved):

PR #41, opened 2020-04-02
Add DiskMappings for specific disk mappings

masayag:
DiskMappings will allow to map source VM disk to storage class.
A mapping for a disk overrides the mapping from  StorageMappings.

DiskMappings will fail to map the disk in the following cases:
* Disk ID/Name doesn't exist for the source VM (ID precedes name)
* Target storage class doesn't exist

DiskMappings.Source.Name should hold the disk alias.

DiskMappings.Source.ID should hold the ID of the disk and not the ID of
the disk-attachment that describes its association to the VM.

DiskMappings is optional, if not provided mappings will rely on
StorageMappings.

### DIFF
--- a/pkg/providers/ovirt/validation/validators/storage-mapping-validator.go
+++ b/pkg/providers/ovirt/validation/validators/storage-mapping-validator.go
@@ -27,34 +27,66 @@ func NewStorageMappingValidator(provider StorageClassProvider) StorageMappingVal
 	}
 }
 
-// ValidateStorageMapping validates storage domain mapping
-func (v *StorageMappingValidator) ValidateStorageMapping(attachments []*ovirtsdk.DiskAttachment, mapping *[]v2vv1alpha1.ResourceMappingItem) []ValidationFailure {
+// ValidateStorageMapping validates storage domain mapping and disk mapping
+func (v *StorageMappingValidator) ValidateStorageMapping(
+	attachments []*ovirtsdk.DiskAttachment,
+	storageMapping *[]v2vv1alpha1.ResourceMappingItem,
+	diskMapping *[]v2vv1alpha1.ResourceMappingItem,
+) []ValidationFailure {
 	var failures []ValidationFailure
 	// Check whether mapping for storage is required and was provided
-	if mapping == nil && len(attachments) > 0 {
+	if storageMapping == nil && diskMapping == nil && len(attachments) > 0 {
 		failures = append(failures, ValidationFailure{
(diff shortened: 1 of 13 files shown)


pkg/providers/ovirt/validation/vm-import-validation.go
```diff
@@ -90,6 +90,11 @@ var checkToAction = map[validators.CheckID]action{
 	validators.VMWatchdogsID:                     block,
 	validators.VMCdromsID:                        log,
 	validators.VMFloppiesID:                      log,
+	// Mapping rules
+	validators.StorageMappingID: block,
```
jakub-dzon: From code perspective it is not needed - the mapping is used only while processing rules validation. Did you add it just for visibility?

pkliczewski: The changes looks good to me. Please check this comment.

masayag: 'block' mappings removed Removed.

jakub-dzon (approved):

pkliczewski (approved):

pkliczewski:
@machacekondra what do you think?

machacekondra (approved):

pkg/providers/ovirt/validation/validators/storage-mapping-validator.go
```diff
@@ -27,34 +27,66 @@ func NewStorageMappingValidator(provider StorageClassProvider) StorageMappingVal
...
+// ValidateStorageMapping validates storage domain mapping and disk mapping
+func (v *StorageMappingValidator) ValidateStorageMapping(
+	attachments []*ovirtsdk.DiskAttachment,
+	storageMapping *[]v2vv1alpha1.ResourceMappingItem,
+	diskMapping *[]v2vv1alpha1.ResourceMappingItem,
+) []ValidationFailure {
 	var failures []ValidationFailure
 	// Check whether mapping for storage is required and was provided
-	if mapping == nil && len(attachments) > 0 {
+	if storageMapping == nil && diskMapping == nil && len(attachments) > 0 {
```
atiratree, line 38: is it really that bad to have both missing? I guess, it can still be used to send empty mapping if the user doesn't want to specify storage classes (or depend on the default one).

It could be just more convenient to send it without the mapping

pkliczewski: I think it is better not to send it if is empty.

PR #42, opened 2020-04-03
Rename and start VM methods

jakub-dzon:
This PR introduces new ovirt client methods:
- for requesting VM startup;
- for changing VM name;

### DIFF
--- a/pkg/providers/ovirt/client/ovirt-client.go
+++ b/pkg/providers/ovirt/client/ovirt-client.go
@@ -18,6 +18,8 @@ const (
 type OvirtClient interface {
 	GetVM(id *string, name *string, cluster *string, clusterID *string) (*ovirtsdk.Vm, error)
 	StopVM(id string) error
+	StartVM(id string) error
+	RenameVM(id string, name string) error
 	Close() error
 }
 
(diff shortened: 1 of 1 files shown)


pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -151,6 +153,44 @@ func (client *richOvirtClient) StopVM(id string) error {
...
+func (client *richOvirtClient) RenameVM(id string, name string) error {
+	vmService := client.connection.SystemService().VmsService().VmService(id)
+
+	vmResponse, _ := vmService.Get().Send()
+	vm, vmAvailable := vmResponse.Vm()
+	if !vmAvailable {
+		return fmt.Errorf("Failed to rename vm %s, vm is not available", id)
+	}
+
+	vm.SetName(name)
```
machacekondra: This is not recomended way of updating vm in oVirt API. The comended way is to just send the parameters you want to update not every parameter. So would be better to change it to something like:

```
newVm, err := ovirtsdk.NewVmBuilder().Name(name).Build()
_, err := vmService.Update().Vm(newVm).Send()
```

jakub-dzon: /release-note-none

jakub-dzon: Thanks!

pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -151,6 +153,44 @@ func (client *richOvirtClient) StopVM(id string) error {
...
+	vmService := client.connection.SystemService().VmsService().VmService(id)
+
+	vmResponse, _ := vmService.Get().Send()
+	vm, vmAvailable := vmResponse.Vm()
+	if !vmAvailable {
+		return fmt.Errorf("Failed to start vm %s, vm is not available", id)
+	}
+
+	// Request the VM startup if VM is not UP
+	if status, _ := vm.Status(); status != ovirtsdk.VMSTATUS_UP {
```
machacekondra: If VM's state is `powering_up` for example. The call to start VM will fail, which is something we don't need I guess. Thinking of that, I think we need to have also some check/validation for HA VM's because we can't stope those.

jakub-dzon: So in case of HA VMs we cannot perform the migration at all?

jakub-dzon: It seems that starting from down state is the safest approach.

machacekondra: Ok, so it seems if it's manual shut-down[0], it's valid:
```
VMs manually shut down or powered off from oVirt, from the guest operating system, or by admin-issued host shutdown (which counts as user initiated shutdown) are not restarted automatically. Switching VMs off manually is a user action and oVirt honors it. The user is responsible for starting such a VM again manually.
```
[0] https://www.ovirt.org/develop/ha-vms.html

machacekondra: Yes, for start status down, is OK I think.

machacekondra (approved):

pkliczewski (approved):

PR #43, opened 2020-04-03
Removed RenameVM method

jakub-dzon:
```release-note
NONE
```


machacekondra (approved):

machacekondra:
Maybe as part of this PR, let's remove also:

https://github.com/kubevirt/vm-import-operator/blob/master/pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go#L176

So we get rid of all UpdateVM stuff.

machacekondra:
Or actually we can still use it for stopVM "updating" status. So let's keep, but just change the comment.

jakub-dzon:
> Or actually we can still use it for stopVM "updating" status. So let's keep, but just change the comment.

Done

pkliczewski (approved):

PR #44, opened 2020-04-03
VM status validation

jakub-dzon:
Allowing only 'up' or 'down'
```release-note
NONE
```

### DIFF
--- a/pkg/providers/ovirt/validation/validators/vm-validator.go
+++ b/pkg/providers/ovirt/validation/validators/vm-validator.go
@@ -11,6 +11,9 @@ import (
 // ValidateVM validates given VM
 func ValidateVM(vm *ovirtsdk.Vm) []ValidationFailure {
 	var results = isValidBios(vm)
+	if failure, valid := isValidStatus(vm); !valid {
+		results = append(results, failure)
+	}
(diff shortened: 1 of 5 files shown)


pkg/providers/ovirt/validation/validators/vm-validator.go
```diff
@@ -93,6 +96,22 @@ func isValidBios(vm *ovirtsdk.Vm) []ValidationFailure {
 	return results
 }
 
+func isValidStatus(vm *ovirtsdk.Vm) (ValidationFailure, bool) {
+	if status, ok := vm.Status(); ok {
+		if status != ovirtsdk.VMSTATUS_UP && status != ovirtsdk.VMSTATUS_DOWN {
+			return ValidationFailure{
+				ID:      VMStatusID,
+				Message: fmt.Sprintf("VM has illegal status: %v", status),
```
machacekondra: Can you please write here, what the valid statuses are? So user know.

jakub-dzon: Good point. Thanks!

jakub-dzon: Done.

pkliczewski (approved):

PR #45, opened 2020-04-03
Using common templates

pkliczewski:
Based on source vm os, workload we need to select a template to be
used for target vm. In case of guest os availble we map oses using
osInfo map. When we only have os - human redable form we attempt to
find which default os to use.

We use medium flavor since we use source vm to update the metadata
and it not important to pick amount of resources used in the template.


pkliczewski:
@suomiy please review

pkliczewski:
@michalskrivanek suggested to take a look at ovirt list of OSes by calling `/ovirt-engine/api/operatingsystems` endpoint.

@masayag suggested to keep os mappings in a `configmap` which would detach mappings from the code so user could update them. We could provide the default configmap together with the operator.

pkg/providers/ovirt/templates/template-finder.go
```diff
@@ -0,0 +1,86 @@
...
+	workload := getWorkload(vm)
+	return f.getTemplate(os, workload)
+}
+
+func findOperatingSystem(vm *ovirtsdk.Vm) string {
+	if gos, found := vm.GuestOperatingSystem(); found {
+		distribution, _ := gos.Distribution()
+		version, _ := gos.Version()
+		fullVersion, _ := version.FullVersion()
+		return fmt.Sprintf("%s%s", osInfo[distribution], fullVersion)
```
atiratree: is the `fullVersion` always available? Can it be also a string incompatible with `osinfo-query`?

pkg/providers/ovirt/templates/template-finder.go
```diff
@@ -0,0 +1,86 @@
...
+		// limit number of possibilities
+		osType = strings.ToLower(osType)
+		if strings.Contains(osType, "linux") || strings.Contains(osType, "rhel") {
+			return defaultLinux
+		} else if strings.Contains(osType, "win") {
+			return defaultWindows
+		}
+	}
+	// return empty to fail label selector
+	return ""
```
atiratree: Shouln't we have a defaultLinux as a fallback? As we have to choose one template?

pkg/providers/ovirt/templates/template-finder.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func getWorkload(vm *ovirtsdk.Vm) string {
+	// vm type is always available
+	vmType, _ := vm.Type()
+	// we need to remove underscore from high_performance, other workloads are OK
+	return strings.Replace(string(vmType), "_", "", -1)
```
atiratree, line 97: is the workload always present? Should we have a default as well?

pkg/providers/ovirt/templates/template-finder.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (f *TemplateFinder) getTemplate(os string, workload string) (*templatev1.Template, error) {
+	// We update metadata from the source vm so we default to medium flavor
+	labelSelector := fmt.Sprintf("os.template.kubevirt.io/%s=true,workload.template.kubevirt.io/%s=true,flavor.template.kubevirt.io/medium=true", os, workload)
```
atiratree: I know it is not necessary now, but we could also try large and other flavors if the medium one is missing. E.g. we are using a customized sort in the UI for that.

pkg/providers/ovirt/templates/template-finder.go
pkliczewski: My understanding is that it is always available. It is just a version and yes it could be incompatible with `os-info`. oVirt do not expose `os-info` so it is the best case we can have.

pkg/providers/ovirt/templates/template-finder.go
pkliczewski: We always will have either `guestOS` or `OS`.  It is here to make compiler happy.

pkg/providers/ovirt/templates/template-finder.go
pkliczewski: yes. No need for default. In oVirt we have: `server`, `desktop` and `high_performance`. The last one needs to be made compatible with KubeVirt where we use `highperformance`.

pkg/providers/ovirt/templates/template-finder.go
pkliczewski: We will update the values anyway so it is transparent what we use. Both `large` and `medium` are for os types in common templates. It was just arbitrary decision to use `medium`.

pkg/providers/ovirt/templates/template-handler.go
```diff
@@ -25,8 +25,8 @@ const (
 	templateNamespaceLabel = "vm.kubevirt.io/template-namespace"
 )
 
-// TemplateFinder attempts to find a template based on given parameters
-type TemplateFinder struct {
+// TemplateHandler attempts to find a template based on given parameters
```
masayag: pls add a notation about the processing part of it

pkliczewski: ack

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -193,7 +201,19 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 
 	// Define VM spec
 	mapper := provider.CreateMapper()
-	vmSpec := mapper.MapVM(instance.Spec.TargetVMName)
+	template, err := provider.FindTemplate()
```
machacekondra, line 204: The only thing we need to improve later is , that we shouldn't process the template for each reconcile. It's not neccessary.

machacekondra (approved):

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
masayag: I think that once the VM is created, there is no need to go through the entire steps.
Perhaps we should store on VMImport CR's annotation the identity of the created VM (namespace/name) and skip all of these steps (validation/mapping/templates/...) since we won't re-create the VM. 

Edited: according to the design, we should store the VM's identity on the CR's status. We can read VM's identity from there and skip all of the unneeded steps if VM was already created.

machacekondra: What you mean by identity? The targetVmName?

pkg/providers/ovirt/provider.go
```diff
@@ -169,6 +177,16 @@ func (o *OvirtProvider) StopVM() error {
 	return nil
 }
 
+// FindTemplate attempts to find best match for a template based on the source VM
+func (o *OvirtProvider) FindTemplate() (*templatev1.Template, error) {
```
jakub-dzon, line 181: Will templates be used only for oVirt? Does it have to be in the provider API? It looks to be provider-agnostic and potentially movable to the controller.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: That will work, I'll refactor the loop as part of #51

pkg/providers/ovirt/templates/os-mapper-provider.go
```diff
@@ -0,0 +1,61 @@
...
+const (
+	osConfigMapName    = "vmimport-os-mapper"
+	guestOsToCommonKey = "guestos2common"
+	osInfoToCommonKey  = "osinfo2common"
+)
+
+// OSMaps is responsible for getting the operating systems maps which contain mapping of GuestOS to common templates
+// and mapping of osinfo to common templates
+type OSMaps struct {
+	Client client.Client
```
jakub-dzon, line 22: Is it likely that templates will be used be used by future providers? 
This looks to be provider-agnostic; what do you think about moving it to pkg/templates?

pkg/providers/ovirt/templates/template-finder.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	if len(templates.Items) == 0 {
+		return nil, fmt.Errorf("Template not found for %s OS and %s workload", os, workload)
+	}
+	// Take first which matches label selector
+	return &templates.Items[0], nil
+}
+
+// ProcessTemplate processes template with provided parameter values
+func (f *TemplateFinder) ProcessTemplate(template *templatev1.Template, vmName string) (*kubevirtv1.VirtualMachine, error) {
```
jakub-dzon, line 116: This also looks like a good candidate for being moved to a provider-agnostic package.

pkg/providers/ovirt/templates/template-provider.go
```diff
@@ -0,0 +1,89 @@
...
+	// TemplateFlavorLabel is a label that specifies the flavor of the template
+	TemplateFlavorLabel = "flavor.template.kubevirt.io/%s"
+
+	processingURI = "processedTemplates"
+	nameParameter = "NAME"
+	otherValue    = "other"
+)
+
+// Templates is responsible for finding templates
+type Templates struct {
```
jakub-dzon, line 29: This one also looks to be something that can be pulled out of ovirt provider scope.

jakub-dzon (commented):
Looks good!:+1: 

Few questions to consider.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: :+1:

pkg/providers/ovirt/provider.go
pkliczewski: It shouldn't be oVirt specific and should be provider agnostic. ATM is part of ovirt provider be we can move it outside in different patch.

pkg/providers/ovirt/templates/os-mapper-provider.go
pkliczewski: :+1:

pkg/providers/ovirt/templates/template-finder.go
pkliczewski: Ack

pkg/providers/ovirt/templates/template-provider.go
pkliczewski: :+1:

jakub-dzon (approved):

PR #46, opened 2020-04-06
Import failure clean-up method

jakub-dzon:
This PR introduces import failure handling logic that should be used in  every place a terminal import failure is detected:
- storing initial source VM status in an annotation;
- re-starting VM if needed;
- removing config map;
- removing secret

```release-note
NONE
```

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -29,8 +29,9 @@ import (
 )
 
 const (
-	keyAccess = "accessKeyId"
-	keySecret = "secretKey"
+	keyAccess            = "accessKeyId"
+	keySecret            = "secretKey"
+	sourceVMInitialState = "vmimport.v2v.kubevirt.io/source-vm-initial-state"
 )
 
 var (
@@ -168,6 +169,17 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 		}
 		if valid, message := shouldFailWith(conditions); !valid {
 			return reconcile.Result{}, langerr.New(message)
+		}
+
+		vmStatus, err := provider.GetVMStatus()
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		reqLogger.Info("Storing source VM status", "status", vmStatus)
+		err = r.storeSourceVMStatus(instance, string(vmStatus))
+		if err != nil {
+			return reconcile.Result{}, err
 		}
 	} else {
 		reqLogger.Info("VirtualMachineImport has already been validated positively. Skipping re-validation")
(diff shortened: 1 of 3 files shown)


pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -29,8 +29,9 @@ import (
 )
 
 const (
-	keyAccess = "accessKeyId"
-	keySecret = "secretKey"
+	keyAccess            = "accessKeyId"
+	keySecret            = "secretKey"
+	sourceVMInitialState = "sourceVmInitialState"
```
masayag: according to API conventions annotations should be prefixed with a label, and prefer teh following format:  `vmimport.v2v.kubevirt.io/source-vm-initial-state`

see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -358,6 +370,84 @@ func (r *ReconcileVirtualMachineImport) upsertStatusConditions(vmiName types.Nam
...
+}
+
+func (r *ReconcileVirtualMachineImport) restoreInitialVMState(vmiName types.NamespacedName, p provider.Provider) error {
+	var instance v2vv1alpha1.VirtualMachineImport
+	err := r.client.Get(context.TODO(), vmiName, &instance)
+	if err != nil {
+		return err
+	}
+	if vmInitialState, ok := instance.Annotations[sourceVMInitialState]; ok {
+		if vmInitialState == string(provider.VMStatusUp) {
```
masayag: you can reduce one nesting level of if if you return quickly in case `ok` is false

pkg/providers/ovirt/provider.go
```diff
@@ -58,6 +58,18 @@ func NewOvirtProvider(vmiCrName types.NamespacedName, client client.Client, kube
...
+func (o *OvirtProvider) GetVMStatus() (provider.VMStatus, error) {
+	if status, ok := o.vm.Status(); ok {
+		switch status {
+		case ovirtsdk.VMSTATUS_DOWN:
+			return provider.VMStatusDown, nil
+		case ovirtsdk.VMSTATUS_UP:
+			return provider.VMStatusUp, nil
+		}
+	}
+	return "", fmt.Errorf("VM doesn't have a legal status")
```
masayag: pls include the illegal status in the message

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
jakub-dzon: Thanks!

jakub-dzon: Fixed.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
jakub-dzon: Good idea. Fixed.

pkg/providers/ovirt/provider.go
jakub-dzon: Done.

machacekondra (approved):

masayag (approved):

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -358,6 +370,86 @@ func (r *ReconcileVirtualMachineImport) upsertStatusConditions(vmiName types.Nam
 	return nil
 }
 
+func (r *ReconcileVirtualMachineImport) storeSourceVMStatus(instance *v2vv1alpha1.VirtualMachineImport, vmStatus string) error {
+	vmiCopy := instance.DeepCopy()
+	vmiCopy.Annotations[sourceVMInitialState] = vmStatus
```
machacekondra, line 375: This is failing on:
```
panic: assignment to entry in nil map
```
we need to init the map before using it.

jakub-dzon: Interesting... For some reason, in my deployment it wasn't empty. Good catch!

PR #47, opened 2020-04-06
After success method

jakub-dzon:
This PR provides method to be executed after a successful VM import.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -381,6 +381,42 @@ func (r *ReconcileVirtualMachineImport) storeSourceVMStatus(instance *v2vv1alpha
 	return r.client.Patch(context.TODO(), vmiCopy, patch)
 }
 
+//TODO: use in proper place
+func (r *ReconcileVirtualMachineImport) afterSuccess(vmName types.NamespacedName, vmiName types.NamespacedName, p provider.Provider) error {
+	var errs []error
+	err := r.afterImport(p, vmiName.Namespace)
+	if err != nil {
+		errs = append(errs, err)
+	}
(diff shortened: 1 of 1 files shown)


pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -381,6 +381,42 @@ func (r *ReconcileVirtualMachineImport) storeSourceVMStatus(instance *v2vv1alpha
...
 }
 
+//TODO: use in proper place
+func (r *ReconcileVirtualMachineImport) afterSuccess(vmName types.NamespacedName, vmiName types.NamespacedName, p provider.Provider) error {
+	var errs []error
+	err := r.afterImport(p, vmiName.Namespace)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	err = r.removeOwnerReference(vmName)
```
pkliczewski: We need to remove `ownerReference` for data volumes, configMaps and secrets.

jakub-dzon: The temporary map and secret are deleted (`afterImport` call) and for dv we don't set the owner reference (at least not yet). I will add the reference removal, though.

machacekondra: Owner references are set in this PR for dv: https://github.com/kubevirt/vm-import-operator/pull/51 so would be nice if you add also cleanup of owner references for DV

jakub-dzon: Will do!

jakub-dzon: Done.

pkliczewski (approved):

machacekondra (approved):

PR #51, opened 2020-04-07
Add watching of VM and DVs

machacekondra:
This PR add:
 - vm import process monitoring as annotation
 - vm start if requested by user
 - set targetVmName of vm in status VMImport CR
 - set datavolumes of vm in status VMImport CR
 - update condtions as per flow of Vmimport process

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -2,8 +2,8 @@ package virtualmachineimport
 
 import (
 	"context"
-	langerr "errors"
 	"fmt"
+	"strconv"
 
 	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
 	"github.com/kubevirt/vm-import-operator/pkg/conditions"
@@ -19,7 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	kubevirtv1 "kubevirt.io/client-go/api/v1"
 	"kubevirt.io/client-go/kubecli"
-
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -32,6 +32,15 @@ import (
 
 const (
 	sourceVMInitialState = "vmimport.v2v.kubevirt.io/source-vm-initial-state"
+	// AnnCurrentProgress is annotations storing current progress of the vm import
+	AnnCurrentProgress = "vmimport.v2v.kubevirt.io/progress"
+	// constants
+	progressStart        = "0"
+	progressCreatingVM   = "30"
+	progressCopyingDisks = "40"
+	progressStartVM      = "90"
+	progressDone         = "100"
+	progressForCopyDisk  = 40
 )
 
 var (
@@ -75,17 +84,34 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	// Watch for changes to primary resource VirtualMachineImport
-	err = c.Watch(&source.Kind{Type: &v2vv1alpha1.VirtualMachineImport{}}, &handler.EnqueueRequestForObject{})
+	err = c.Watch(
+		&source.Kind{Type: &v2vv1alpha1.VirtualMachineImport{}},
+		&handler.EnqueueRequestForObject{},
+	)
 	if err != nil {
 		return err
 	}
 
-	// TODO(user): Modify this to be the types you create that are owned by the primary resource
-	// Watch for changes to secondary resource Pods and requeue the owner VirtualMachineImport
-	err = c.Watch(&source.Kind{Type: &kubevirtv1.VirtualMachine{}}, &handler.EnqueueRequestForOwner{
-		IsController: true,
-		OwnerType:    &v2vv1alpha1.VirtualMachineImport{},
-	})
+	// Watch for VM events:
+	err = c.Watch(
+		&source.Kind{Type: &kubevirtv1.VirtualMachine{}},
+		&handler.EnqueueRequestForOwner{
+			IsController: true,
+			OwnerType:    &v2vv1alpha1.VirtualMachineImport{},
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	// Watch for DV events:
+	err = c.Watch(
+		&source.Kind{Type: &cdiv1.DataVolume{}},
+		&handler.EnqueueRequestForOwner{
+			IsController: true,
+			OwnerType:    &v2vv1alpha1.VirtualMachineImport{},
+		},
+	)
 	if err != nil {
 		return err
 	}
@@ -113,9 +139,6 @@ type ReconcileVirtualMachineImport struct {
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
-	reqLogger.Info("Reconciling VirtualMachineImport")
-
 	// Fetch the VirtualMachineImport instance
 	instance := &v2vv1alpha1.VirtualMachineImport{}
 	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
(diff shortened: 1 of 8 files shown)


pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -199,6 +201,9 @@ func (o *OvirtMapper) mapNics() []kubevirtv1.Interface {
 		if nicInterface, ok := nic.Interface(); ok {
 			kubevirtNic.Model = string(nicInterface)
 		}
+		// FIXME: should this be as part of mapping?
+		// options: slirp, bridge, sriov and masquerade
```
pkliczewski: VM migration works only for `masquerade`. @phoracek any suggestions?

phoracek: Any context on what's going on here?

pkliczewski: This code imports vms from ovirt to kubevirt. The question is about which option to use.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -200,59 +218,147 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+	// Import disks:
+	dvs := mapper.MapDisks()
+	dvsDone := make(map[string]bool)
+	for dvID := range dvs {
+		foundDv, err := r.kubeClient.CdiClient().CdiV1alpha1().DataVolumes(instance.Namespace).Get(dvID, metav1.GetOptions{})
+		if err != nil && errors.IsNotFound(err) {
+			// Create it:
+			// Secret with username/password for the image import:
+			dvCreds := provider.GetDataVolumeCredentials()
+			if err = r.ensureDVSecretExists(instance, vmSpec.Namespace, dvCreds); err != nil {
```
jakub-dzon: Why in the loop? If there are multiple dvs, that method will be called for each of them in the first reconciliation (and that entails multiple secret retrievals). 
Of course, it will be called zero times if the dvs are already being imported vs. always once outside of the loop.:)

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -200,59 +218,147 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+				return reconcile.Result{}, err
+			}
 
-		// Import disks:
-		dvs := mapper.MapDisks()
-		for _, dv := range dvs {
-			_, err := r.kubeClient.CdiClient().CdiV1alpha1().DataVolumes(instance.Namespace).Create(&dv)
+			// CM containing CA for the image import:
+			if err = r.ensureDVConfigMapExists(instance, vmSpec.Namespace, dvCreds); err != nil {
```
jakub-dzon: Same here

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Well, right. I think better place will be to call it, in VM creation block, because it's called only once.

pkg/providers/ovirt/mapper/mapper.go
phoracek: If you want to be always able to import oVirt VM to KubeVirt, you should just give it a single network interface that would be connected through `masquerade` to the pod network.

If you really want to give the VM the same environment as it had on oVirt, somebody needs to configure the L2 network (possibly via kubernetes-nmstate with bridges representing each oVirt network) and configure NetworkAttachmentDefiniton per each. Then you would add one network interface with `bridge` binding connected to respective multus network.

pkliczewski: We do not want to configure anything. We expect it to be preconfigured, if it is not there we should fail the process. As I understand we should use `masquerade` for pod network and `bridge` for multus networks. Please correct me if I am wrong.

phoracek: That's right. So what you can do is to add Interface to the VM which would be based on a Multus network and would use bridge binding.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -32,6 +32,17 @@ import (
 
 const (
 	sourceVMInitialState = "vmimport.v2v.kubevirt.io/source-vm-initial-state"
+	// AnnCurrentProgress is annotations storing current progress of the vm import
+	AnnCurrentProgress = "vmimport.v2v.kubevirt.io/progress"
+	// constants
+	progressStart            = "0"
+	progressValidation       = "10"
```
jakub-dzon: unused

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -32,6 +32,17 @@ import (
...
+	progressUpdatingSourceVM = "20"
```
jakub-dzon: unused

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -503,3 +685,72 @@ func shouldFailWith(conditions []v2vv1alpha1.VirtualMachineImportCondition) (boo
...
+	provider, err := r.createProvider(instance)
+	if err != nil {
+		return nil, err
+	}
+
+	err = provider.Connect(sourceProviderSecretObj)
+	if err != nil {
+		return nil, err
+	}
+	defer provider.Close()
```
jakub-dzon: It will disconnect when the function ends; we will need the ovirt client connected later.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -130,139 +158,207 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
-		conditions, err := provider.Validate()
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-		err = r.upsertStatusConditions(request.NamespacedName, conditions)
+	vmName := types.NamespacedName{Name: instance.Status.TargetVMName, Namespace: request.Namespace}
+	if instance.Status.TargetVMName == "" {
+		// Define VM spec
+		template, err := provider.FindTemplate()
+		var spec = &kubevirtv1.VirtualMachine{}
```
jakub-dzon: Spec is always assigned later; no need to initialize it here explicitly.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -130,139 +158,207 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
 			return reconcile.Result{}, err
 		}
 
-		// Pod created successfully - don't requeue
-		return reconcile.Result{}, nil
+		// Update pogress to creating vm
+		if err = r.updateProgress(instance, progressCreatingVM); err != nil {
+			return reconcile.Result{}, err
+		}
 	} else if err != nil {
```
jakub-dzon: `err` will always be nil- that branch can be removed

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -32,6 +32,17 @@ import (
...
+	AnnCurrentProgress = "vmimport.v2v.kubevirt.io/progress"
+	// constants
+	progressStart            = "0"
+	progressValidation       = "10"
+	progressUpdatingSourceVM = "20"
+	progressCreatingVM       = "30"
+	progressCopyingDisks     = "40"
+	progressStartVM          = "90"
+	progressDone             = "100"
+	progressForCopyDisk      = 40
```
pkliczewski: I see here `progressForCopyDisk` and `progressCopyingDisks`. One with int and the other with string as value. Do we need both?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -130,139 +158,207 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
 	}
 
-	// VM already exists - don't requeue
-	reqLogger.Info("Skip reconcile: VM already exists", "VM.Namespace", found.Namespace, "VM.Name", found.Name)
+	// Import disks:
+	dvs := mapper.MapDisks()
+	dvsDone := make(map[string]bool)
+	for dvID := range dvs {
+		foundDv, err := r.kubeClient.CdiClient().CdiV1alpha1().DataVolumes(instance.Namespace).Get(dvID, metav1.GetOptions{})
+		if err != nil && errors.IsNotFound(err) {
```
jakub-dzon, line 183: Can that code be extracted to a separate method/function?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -130,139 +158,207 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+			provider.UpdateVM(vmDef, dvs)
+			err = r.client.Update(context.TODO(), vmDef)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		} else if err == nil {
+			// Check if all DVs was imported:
```
jakub-dzon: Can that code be extracted to a separate method/function?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -130,139 +158,207 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+					}
+				}
+			}
+		} else {
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Start vm
+	if *instance.Spec.StartVM && conditions.HasSucceededConditionOfReason(instance.Status.Conditions, v2vv1alpha1.VirtualMachineReady) {
```
jakub-dzon, line 291: Can that code be extracted to a separate method/function?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -503,3 +685,72 @@ func shouldFailWith(conditions []v2vv1alpha1.VirtualMachineImportCondition) (boo
 	}
 	return valid, message
 }
+
+func (r *ReconcileVirtualMachineImport) initProvider(instance *v2vv1alpha1.VirtualMachineImport) (provider.Provider, error) {
```
jakub-dzon, line 726: :+1:

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -130,139 +158,207 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+			}
+			for _, dv := range dvs {
+				if err := controllerutil.SetControllerReference(instance, &dv, r.scheme); err != nil {
+					return reconcile.Result{}, err
+				}
+				_, err := r.kubeClient.CdiClient().CdiV1alpha1().DataVolumes(instance.Namespace).Create(&dv)
+				if err != nil {
+					// Update condtion
+					cond = conditions.NewSucceededCondition(string(v2vv1alpha1.DataVolumeCreationFailed), fmt.Sprintf("Data volume creationg faield: %s", err))
+					err = r.upsertStatusCondition(request.NamespacedName, cond)
```
jakub-dzon: `err` is ignored; can be `_ =`

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -130,139 +158,207 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
-
-	// Validate if it's needed at this stage of processing
-	if shouldValidate(&instance.Status) {
-		conditions, err := provider.Validate()
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-		err = r.upsertStatusConditions(request.NamespacedName, conditions)
+	vmName := types.NamespacedName{Name: instance.Status.TargetVMName, Namespace: request.Namespace}
+	if instance.Status.TargetVMName == "" {
```
pkliczewski, line 170: The logic looks good but I wonder whether it would make sense to split it in to bunch of methods in the next PR. This would allow us to write some tests.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -503,3 +685,72 @@ func shouldFailWith(conditions []v2vv1alpha1.VirtualMachineImportCondition) (boo
...
+	// Load source VM:
+	err = provider.LoadVM(instance.Spec.Source)
+	if err != nil {
+		return nil, err
+	}
+
+	// Load the external resource mapping
+	resourceMapping, err := r.fetchResourceMapping(instance.Spec.ResourceMapping, instance.Namespace)
+	if err != nil {
+		//TODO: update Validating status condition
```
pkliczewski, line 752: Do we want to update status condition in this PR or next one?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -130,139 +158,207 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+	for dvID := range dvs {
+		foundDv, err := r.kubeClient.CdiClient().CdiV1alpha1().DataVolumes(instance.Namespace).Get(dvID, metav1.GetOptions{})
+		if err != nil && errors.IsNotFound(err) {
+			// Update condition to create VM:
+			cond := conditions.NewProccessingCondition(string(v2vv1alpha1.CopyingDisks), "Copying virtual machine disks")
+			err = r.upsertStatusCondition(request.NamespacedName, cond)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+			// Update pogress to done
```
jakub-dzon: "progress"

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -130,139 +158,207 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+			if err = r.updateProgress(instance, progressCopyingDisks); err != nil {
+				return reconcile.Result{}, err
+			}
+			for _, dv := range dvs {
+				if err := controllerutil.SetControllerReference(instance, &dv, r.scheme); err != nil {
+					return reconcile.Result{}, err
+				}
+				_, err := r.kubeClient.CdiClient().CdiV1alpha1().DataVolumes(instance.Namespace).Create(&dv)
+				if err != nil {
+					// Update condtion
```
jakub-dzon: "condition"

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -130,139 +158,207 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+		vmName.Name = found.Name
+
+		// Set target name:
+		if err = r.updateTargetVMName(request.NamespacedName, vmName.Name); err != nil {
 			return reconcile.Result{}, err
 		}
 
-		// Pod created successfully - don't requeue
-		return reconcile.Result{}, nil
+		// Update pogress to creating vm
```
jakub-dzon: "progress"

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: `progressForCopyDisk` is used to calculate the progress from the number of disks as:
`progressCopyingDisks + (progressForCopyDisk/numberofdisksDone)`

machacekondra: So when all disks are done we are at 80% and when we start copying disk we are at 40%. When there are two disks, and first is done we are at 60%. When both we are at 80%

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Sure

jakub-dzon (changes requested):
General comment to the `Reconcile` method: it would be much easier to reason about it if it was split into smaller methods. 
Current comments seem to mark good candidates for extraction (i.e. `Start vm`,  `Import disks:`, `If all DVs was imported update state, and if VM shouldn't be started end the import:`, etc.).

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Actually it is, because it's used outside if-else block.

jakub-dzon: Yes, and it is assigned in every path of execution either from `ProcessTemplate` or `CreateEmptyVM`.

`var spec = &kubevirtv1.VirtualMachine{}` creates `VirtualMachine` and then assigns pointer to it to `spec`
vs. 
`var spec *kubevirtv1.VirtualMachine` is initialized with pointer's zero value - `nil`
And the `spec` pointer is always overwritten in if-else.

machacekondra: Ah, I thought to declare it, sorry :) Fixed.

machacekondra:
Thanks for review. I've fixed your comments.

jakub-dzon (approved):
:+1:

pkliczewski (approved):

PR #54, opened 2020-04-08
Data Volume Secret and Config Map management refactoring

jakub-dzon:
Changes introduced by this PR:
- DV config map and secret management has been moved to oVirt provider, as it is oVirt-specific, which includes:
  - creation;
  - visibility restricted to the provider code;
  - clean-up (removal);
- DV config map and secret names will be generated and the objects will be  labelled with the name of "parent" Virtual Machine Import;
- DV config map and secret will be retrieved from the cluster only using label selectors.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -31,8 +31,6 @@ import (
 )
 
 const (
-	keyAccess            = "accessKeyId"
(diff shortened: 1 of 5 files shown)


pkliczewski (approved):

machacekondra:
Let's merge first the templates, I think this will be easier to rebase.

jakub-dzon:
> Let's merge first the templates, I think this will be easier to rebase.

:+1:

PR #59, opened 2020-04-08
Ignore if VM already exists

machacekondra:
On slow env it may happen, that when we call createVm, and right after that call we fetch it. The VM is not yet in the database. So in this case we need to ignore another call createVM, and go through of the rest of the reconcile process.


machacekondra:
Fixes: https://github.com/kubevirt/vm-import-operator/issues/84

pkliczewski (approved):

Issue #62, opened 2020-04-09
Sometimes the logs are flooded by disk 123 has illegal status: 'locked'. Allowed value: 'ok'


machacekondra:
The disks is imported successfully, so it seems locked is just OK intermediate stat, so we shouldn't flood the log.

```
2020-04-09T12:21:32.601+0200	ERROR	controller-runtime.controller	Reconciler error	{"controller": "virtualmachineimport-controller", "request": "default/myvm", "error": "disk 64302e7f-3f08-4d32-9fe1-59b6c383acb5 has illegal status: 'locked'. Allowed value: 'ok'"}
github.com/go-logr/zapr.(*zapLogger).Error
	/Users/omachace/go/src/github.com/kubevirt/vm-import-operator/vendor/github.com/go-logr/zapr/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/Users/omachace/go/src/github.com/kubevirt/vm-import-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:258
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/Users/omachace/go/src/github.com/kubevirt/vm-import-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:232
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
	/Users/omachace/go/src/github.com/kubevirt/vm-import-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:211
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
	/Users/omachace/go/src/github.com/kubevirt/vm-import-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:152
k8s.io/apimachinery/pkg/util/wait.JitterUntil
	/Users/omachace/go/src/github.com/kubevirt/vm-import-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:153
k8s.io/apimachinery/pkg/util/wait.Until
	/Users/omachace/go/src/github.com/kubevirt/vm-import-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88```

jakub-dzon:
[...]

Issue #61, opened 2020-04-09
Remove Warnings from log


machacekondra:
```
W0409 12:21:38.288715   26671 reflector.go:299] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:204: watch of *v1alpha1.VirtualMachineImport ended with: an error on the server ("unable to decode an event from the watch stream: unable to decode watch event: no kind \"VirtualMachineImport\" is registered for version \"v2v.kubevirt.io/v1alpha1\" in scheme \"kubevirt.io/client-go/api/v1/register.go:78\"") has prevented the request from succeeding
```

pkliczewski:
Do you have any clue when this warning can occur? It seems like we need to register vm-import scheme.

machacekondra:
We have it registered. I am not sure why it happens. I will investigate. It also reports this for other resources.

```
W0409 13:56:40.389441   33597 reflector.go:299] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:204: watch of *v1alpha1.VirtualMachineImport ended with: an error on the server ("unable to decode an event from the watch stream: unable to decode watch event: no kind \"VirtualMachineImport\" is registered for version \"v2v.kubevirt.io/v1alpha1\" in scheme \"kubevirt.io/client-go/api/v1/register.go:78\"") has prevented the request from succeeding
W0409 13:56:40.650406   33597 reflector.go:299] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:204: watch of *v1.Secret ended with: an error on the server ("unable to decode an event from the watch stream: unable to decode watch event: no kind \"Secret\" is registered for version \"v1\" in scheme \"kubevirt.io/client-go/api/v1/register.go:78\"") has prevented the request from succeeding
W0409 13:56:40.658203   33597 reflector.go:299] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:204: watch of *v1.ConfigMap ended with: an error on the server ("unable to decode an event from the watch stream: unable to decode watch event: no kind \"ConfigMap\" is registered for version \"v1\" in scheme \"kubevirt.io/client-go/api/v1/register.go:78\"") has prevented the request from succeeding
[...]

PR #70, opened 2020-04-09
Don't use kubevirt client

machacekondra:
Use internal one instead. It's not recommended to use non-default ones.


machacekondra:
Fixes: #61

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -60,19 +59,14 @@ func Add(mgr manager.Manager) error {
...
-	}
 	tempClient, err := templatev1.NewForConfig(mgr.GetConfig())
 	if err != nil {
 		log.Error(err, "Unable to get OC client")
 		panic("Controller cannot operate without OC client")
 	}
 	client := mgr.GetClient()
 	finder := mappings.NewResourceMappingsFinder(client)
-	return &ReconcileVirtualMachineImport{client: client, scheme: mgr.GetScheme(), kubeClient: kubeClient, resourceMappingsFinder: finder, ocClient: tempClient}
+	return &ReconcileVirtualMachineImport{client: client, scheme: mgr.GetScheme(), resourceMappingsFinder: finder, ocClient: tempClient}
```
pkliczewski, line 69: maybe we could get rid of `ocClient` as well?

machacekondra:
It fixes also #62 . Those errors re-called reconcile and so that's why we saw it. 
Fixes: #62

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: Yes, this PR removes only kubevirt client. I will send another one with oclient

pkliczewski: I meant the last parameter

```
ocClient: tempClient
```

machacekondra: Yes, I know, but it would mean remove it from all places same as kubevirt client. And it's quite lot changes, so I wanted to do this patch smaller.

pkliczewski (approved):

jakub-dzon (approved):
:+1: 
`+60 −33,116`: wonderful!:D

PR #82, opened 2020-04-14
Set VM as owner of DV

### LINKED ISSUE #74: The owner of DataVolumes should be the VM after successful import

machacekondra:
Fixes: https://github.com/kubevirt/vm-import-operator/issues/74

### DIFF
--- a/pkg/ownerreferences/ownerreferences.go
+++ b/pkg/ownerreferences/ownerreferences.go
@@ -1,0 +1,126 @@
+package ownerreferences
+
+import (
+	"context"
+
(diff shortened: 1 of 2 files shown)


pkliczewski (approved):

pkg/ownerreferences/ownerreferences.go
```diff
@@ -0,0 +1,126 @@
+package ownerreferences
```
jakub-dzon, line 1: Good idea to extract that functionality! :+1:

jakub-dzon (approved):

PR #85, opened 2020-04-14
Stopping oVirt VM if cached VM status is not 'down'

### LINKED ISSUE #78: Source vm is always stopped
While working on controller tests I noticed that we always [attempt](https://github.com/kubevirt/vm-import-operator/blob/master/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go#L171) to stop vm without checking its status first

jakub-dzon:
- oVirt client will aways send shutdown command;
 - oVirt provider will request VM shutdown only if locally cached VM status is not 'down';

Fixes #78

```release-note
NONE
```

### DIFF
--- a/pkg/providers/ovirt/client/ovirt-client.go
+++ b/pkg/providers/ovirt/client/ovirt-client.go
@@ -113,28 +113,20 @@ func (client *richOvirtClient) GetVM(id *string, name *string, cluster *string, 
 func (client *richOvirtClient) StopVM(id string) error {
 	vmService := client.connection.SystemService().VmsService().VmService(id)
 
-	// Return if VM is already stopped:
-	vmResponse, _ := vmService.Get().Send()
-	vm, vmAvailable := vmResponse.Vm()
-	if !vmAvailable {
-		return fmt.Errorf("Failed to stop vm %s, vm is not available", id)
-	}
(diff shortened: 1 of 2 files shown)


pkg/providers/ovirt/client/ovirt-client.go
```diff
@@ -148,11 +140,15 @@ func (client *richOvirtClient) StopVM(id string) error {
...
 		if !success {
-			return fmt.Errorf("Failed to stop vm %s", vm.MustName())
+			return fmt.Errorf("Failed to stop vm %s", id)
 		}
 		return nil
 	case <-time.After(vmStopTimeout * time.Minute):
-		return fmt.Errorf("Failed to stop vm %s, current status is %s", vm.MustName(), vm.MustStatus())
+		status := ovirtsdk.VMSTATUS_UNKNOWN
+		if vm != nil {
+			status = vm.MustStatus()
```
pkliczewski: `MustStatus` could cause panic. Can we use `Status` function?

machacekondra (approved):

pkg/providers/ovirt/client/ovirt-client.go
jakub-dzon: Done.

pkliczewski (approved):

machacekondra (approved):

Issue #72, opened 2020-04-09
Memory and Disk sizes in DV should be specified at least in Ki


atiratree:
- or convert to the highest compatible unit

pkliczewski:
@suomiy please add some examples

atiratree:
for example in vm spec
```
        resources:
          limits:
            memory: '536870912'
          requests:
            memory: '536870912'
```

This could read `512 Mi` instead

or the datavolume spec

```
spec:
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: '46137344'
    storageClassName: local
```

it could read `44Mi`

PR #87, opened 2020-04-15
Use suffixes in memory and disk size

machacekondra:
This PR add a suffix to memory and disk attributes, so it is better
readable for the user instead of using bytes.

### DIFF
--- a/pkg/providers/ovirt/mapper/mapper.go
+++ b/pkg/providers/ovirt/mapper/mapper.go
@@ -2,7 +2,6 @@ package mapper
 
 import (
 	"fmt"
-	"strconv"
 	"strings"
 
 	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
@@ -87,7 +86,8 @@ func (o *OvirtMapper) CreateEmptyVM() *kubevirtv1.VirtualMachine {
 }
 
 // MapVM map oVirt API VM definition to kubevirt VM definition
-func (o *OvirtMapper) MapVM(targetVMName *string, vmSpec *kubevirtv1.VirtualMachine) *kubevirtv1.VirtualMachine {
+func (o *OvirtMapper) MapVM(targetVMName *string, vmSpec *kubevirtv1.VirtualMachine) (*kubevirtv1.VirtualMachine, error) {
+	var err error
 	// Set Namespace
 	vmSpec.ObjectMeta.Namespace = o.namespace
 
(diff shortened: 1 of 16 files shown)


machacekondra:
Fixes: https://github.com/kubevirt/vm-import-operator/issues/72

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -164,7 +163,7 @@ func (o *OvirtMapper) MapDisks() map[string]cdiv1.DataVolume {
 		diskID, _ := disk.Id()
 		sdClass := o.getStorageClassForDisk(disk, o.mappings)
 		diskSize, _ := disk.ProvisionedSize()
-		quantity, _ := resource.ParseQuantity(strconv.FormatInt(diskSize, 10))
+		quantity, _ := resource.ParseQuantity(utils.FormatBytes(diskSize))
```
jakub-dzon: mapper tests are failing due to that change.

pkg/utils/utils.go
```diff
@@ -115,3 +116,22 @@ func CountImportedDataVolumes(dvsDone map[string]bool) int {
 
 	return done
 }
+
+// FormatBytes convert bytes to highest suffix
+func FormatBytes(bytes int64) string {
+	if bytes < 0 {
+		return ""
```
jakub-dzon: What do you think abou returning error instead?

pkg/utils/utils.go
```diff
@@ -115,3 +116,22 @@ func CountImportedDataVolumes(dvsDone map[string]bool) int {
...
+	if bytes < kib {
+		return fmt.Sprintf("%d", bytes)
+	}
+	suffix := 0
+	n := int64(kib)
+	for i := bytes / kib; i >= kib; i /= kib {
+		n *= kib
+		suffix++
+	}
+	return fmt.Sprintf("%d%ci", bytes/n, "KMGTPE"[suffix])
```
jakub-dzon: Is it ok to lose precision? It is very visible  with higher numbers i.e. for `1024*1024*1024*1024-1` you get 1023Gi - that is a difference of 1Gi,

jakub-dzon (changes requested):

pkg/utils/utils.go
machacekondra: Well, right. I think I will round it. I check and in ovirt it's not possible to create VM with memory like `1024*1024*1024*1024-1`. It always rounded to MiB. For storage size it's rounded to KiB.

jakub-dzon (approved):

pkliczewski (approved):

PR #88, opened 2020-04-15
test: controller without reconcile

pkliczewski:
This PR provides over 70% test coverage to contoller code.

### DIFF
--- a/pkg/controller/virtualmachineimport/controller_suite_test.go
+++ b/pkg/controller/virtualmachineimport/controller_suite_test.go
@@ -1,0 +1,13 @@
+package virtualmachineimport
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestController(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Controller Suite")
+}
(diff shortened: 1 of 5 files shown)


jakub-dzon (commented):
Looks good!:+1:

What do you think about checking that status conditions are patched properly in all the places - maybe in a separate PR? 
I think it might be helpful to have the whole state machine under control.

pkliczewski:
> Looks good!
> 
> What do you think about checking that status conditions are patched properly in all the places - maybe in a separate PR?
> I think it might be helpful to have the whole state machine under control.

I like the idea. My intention for this PR was to have working mocks and cover as much code as possible. When you check the tests they only check whether the `error` is there or not. We need to add actual logic verification. There should be few more PRs to make sure we are covered.

jakub-dzon (approved):

pkg/controller/virtualmachineimport/controller_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package virtualmachineimport
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestTemplates(t *testing.T) {
```
masayag: Perhaps TestController ?

pkliczewski: :+1:

machacekondra (approved):

Issue #64, opened 2020-04-09
Don't reconnect to target infra if not needed


machacekondra:
It may be very time consuming to connect to target infra, so don't connect there if not needed.

pkliczewski:
Please provide more info when we allow to reconnect?

jakub-dzon:
@machacekondra - can you describe in more detail what you meant?

machacekondra:
We call initProvider each time reconcile is invoked. But we need to call it only once to fetch VM to validate it and map it, later when disks are  being copied, and we only monitor status, it's not needed to call initProvider anymore, it would need some refactoring of the provider module. It would improve performance in case the connection is expensive.

PR #100, opened 2020-04-17
tests: reconcile

pkliczewski:
This PR provides tests for reconcile logic and introduces ability to mock oVirt client.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1,0 +1,14 @@
+package client
+
+// Factory creates new clients
+type Factory interface {
+	NewOvirtClient(dataMap map[string]string) (VMClient, error)
+}
+
+// VMClient provides interface how source virtual machines should be fetched
+type VMClient interface {
+	GetVM(id *string, name *string, cluster *string, clusterID *string) (interface{}, error)
+	StopVM(id string) error
+	StartVM(id string) error
+	Close() error
+}
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
+	aclient "github.com/kubevirt/vm-import-operator/pkg/client"
 	"github.com/kubevirt/vm-import-operator/pkg/conditions"
 	"github.com/kubevirt/vm-import-operator/pkg/mappings"
 	"github.com/kubevirt/vm-import-operator/pkg/ownerreferences"
@@ -69,7 +70,8 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 	client := mgr.GetClient()
 	finder := mappings.NewResourceMappingsFinder(client)
 	ownerreferencesmgr := ownerreferences.NewOwnerReferenceManager(client)
-	return &ReconcileVirtualMachineImport{client: client, scheme: mgr.GetScheme(), resourceMappingsFinder: finder, ocClient: tempClient, ownerreferencesmgr: ownerreferencesmgr}
+	factory := ovirtprovider.NewSourceClientFactory()
+	return &ReconcileVirtualMachineImport{client: client, scheme: mgr.GetScheme(), resourceMappingsFinder: finder, ocClient: tempClient, ownerreferencesmgr: ownerreferencesmgr, factory: factory}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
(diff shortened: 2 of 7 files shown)


pkg/providers/provider.go
```diff
@@ -41,3 +43,21 @@ type Mapper interface {
...
+// SourceClientFactory provides default client factory implementation
+type SourceClientFactory struct{}
+
+// NewSourceClientFactory creates new factory
+func NewSourceClientFactory() *SourceClientFactory {
+	return &SourceClientFactory{}
+}
+
+// NewOvirtClient creates new Ovirt clients
+func (f *SourceClientFactory) NewOvirtClient(dataMap map[string]string) (aclient.VMClient, error) {
```
jakub-dzon: I think it should not be located here - that package should be general and shouldn't know anything about its implementations.

pkg/client/client.go
```diff
@@ -0,0 +1,14 @@
+package client
+
+// Facotry creates new clients
+type Facotry interface {
+	NewOvirtClient(dataMap map[string]string) (VMClient, error)
+}
+
+// VMClient provides interface how source virtual machines should be fetched
+type VMClient interface {
```
jakub-dzon, line 9: I am not sure that it needs to be so generic - other providers might need other operations and the operations are executed only by the concrete providers.

What do you think about creating provider factory, that would assemble providers with what they need instead? Is control over vm client needed in "general" packages? All that can be mocked on the client level can be mocked on provider (from controller perspective). I think vm client should be an implementation detail of each provider.

Generally I think that layer of code should only have access to abstration over its lower level if it needs to be both extensible and loosely-coupled. In this case we give controller access to abstraction over code two levels below.

jakub-dzon (changes requested):

pkg/client/client.go
```diff
@@ -0,0 +1,14 @@
+package client
+
+// Facotry creates new clients
+type Facotry interface {
```
machacekondra: Factory

pkliczewski: will fix.

pkg/providers/provider.go
pkliczewski: Do you suggest to move it to `OvirtProvider`?

pkg/client/client.go
pkliczewski: Interesting idea, will explore.

pkg/providers/provider.go
jakub-dzon: Yes, or to the potential provider factory.

pkg/client/client.go
pkliczewski: I thought about it how to do it well and it seems that provider factory won't bring a lot of value. We still would need a way to detach client and provider life cycles. We want to use the client when it is needed as specified in issue https://github.com/kubevirt/vm-import-operator/issues/64. I will move it to `OvirtProvider`.

jakub-dzon: Do you plan to use it directly from the controller? Won't provider lifecycle methods (maybe split/refactored) be sufficient?

pkliczewski: The idea is to have a factory which we can use to create client. In a unit test we need to have a way to mock the client. Provider methods are not sufficient for it.

jakub-dzon: :+1:

pkliczewski:
@jakub-dzon Please review again

jakub-dzon:
I don't understand why control over ovirt client creation is needed in the controller tests - is the controller going to interact with it directly as a solution to #64? If not, then why not just mock the provider and test ovirt client-related edge cases in ovirt provider tests?

pkliczewski:
The idea is that we want to decouple client and provider life cycle. We do not want to connect when the provider is created and only connect when needed. We could come up with provider factory and mock it like we do for there tests but preparation to fix https://github.com/kubevirt/vm-import-operator/issues/64 needs to be done anyway. This PR helps to solve one problem which helps to fix the other one.

jakub-dzon (approved):

machacekondra (approved):

machacekondra:
@pkliczewski Can you rebase? I will merge then.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
+	aclient "github.com/kubevirt/vm-import-operator/pkg/client"
```
masayag, line 10: just a nit - I'd call it pclient (p for provider), but we can leave it as is

pkliczewski:
@machacekondra please take a look

machacekondra (approved):

masayag (approved):

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: Naming is the hardest thing for me. Will push a PR to change it. Thanks!

PR #113, opened 2020-04-21
client: connect oVirt client only when needed

### LINKED ISSUE #64: Don't reconnect to target infra if not needed

pkliczewski:
In this PR we only use oVirt client when needed. It is connected on demand
whenever there is a need to fetch some data or perform any actions in
source infrastructure.

Fixes: https://github.com/kubevirt/vm-import-operator/issues/64 


masayag (approved):
Looks good to me

jakub-dzon (approved):
It looks good.

Is there still a need for a generic client factory and interface? oVirt client is created and used only by the oVirt provider Plus the controller doesn't need to know about the factory either.

pkliczewski:
@jakub-dzon Yes, we use client factory. In the future we can extend it to make other clients. The interface defines how those clients should look like.

jakub-dzon:
I don't see any place it is used outside of the concrete, oVirt-specific provider - both `VMClient` and `Factory` (besides passing it to the provider "constructor")

Other clients may require different API, tailored to the specific provider. In controller tests provider can be mocked, not the client factory.

pkliczewski:
`Factory` with the client interface is stored in `ReconcileVirtualMachineImport` when we create new reconciler. The idea was that it can be used for other providers in the future. Clients for other providers  can have their own interface(s) but we need our `VMClient` to implement mocks.

pkliczewski:
As outcome of the conversation following issue was created https://github.com/kubevirt/vm-import-operator/issues/125

machacekondra (approved):

PR #151, opened 2020-04-27
Add Makefile for building, testing and releasing of the operator 

### LINKED ISSUE #140: Add manifests folder
The 'deploy' folder needs to be replaced by 'manifests' in which versioning of the operator will be managed.

masayag:
The patch introduces the following changes:

- Add Makefile for building and checking code quality
- Trailing whitespaces are forbidden
- Manifest are created per release in  `manifests/vm-import-operator/$VERSION`
- RBAC resources are squashed into _operator.yaml_
- `make prepare-patch|minor|major` will create a new set of manifests and will bump version of _build|minor|major_ respectively.
- `make docker-build` and `make docker-push` will be used for pushing the built image to _quay.io/kubevirt_ (within this version, target repository is set for $USER due to lack of permissions for _quay.io/kubevirt_)

Fixes #140


.coverprofile
```diff
@@ -0,0 +1,1078 @@
+mode: set
```
jakub-dzon: Shouldn't that file be ignored?

Makefile
```diff
@@ -0,0 +1,135 @@
...
+	CONTAINER_TAG=$(IMAGE_TAG) \
+	REPLACE_KUBEVIRT_NAMESPACE=$(TARGET_NAMESPACE) \
+	OPERATOR_IMAGE=$(OPERATOR_IMAGE) \
+		./hack/generate-manifests.sh
+
+gen-k8s: $(OPERATOR_SDK) $(apis_sources)
+	$(OPERATOR_SDK) generate k8s
+	touch $@
+
+gen-k8s-check: $(apis_sources)
```
jakub-dzon, line 101: What do you think about including clients generation as well (https://github.com/kubevirt/vm-import-operator/blob/master/hack/update-codegen.sh)

machacekondra (approved):

jakub-dzon (commented):
What is the purpose of version/description file?

Makefile
```diff
@@ -0,0 +1,135 @@
+all: fmt check
+
+# Always keep the last released version here
+VERSION ?= 0.0.1
```
pkliczewski, line 4: What about getting version from the latest tag.

Makefile
```diff
@@ -0,0 +1,135 @@
...
+export VERSION := $(VERSION)
+
+TARGET_NAMESPACE ?= kubevirt-hyperconverged
+export TARGET_NAMESPACE := $(TARGET_NAMESPACE)
+
+DEPLOY_DIR ?= manifests
+
+# replace with quay.io/kubevirt once permissions are granted
+IMAGE_REGISTRY ?= quay.io/$(USER)
```
pkliczewski: It still would be good to have an option to override a user for development.

.coverprofile
pkliczewski: :+1:

masayag: no, it shouldn't have been added from the beginning. I'll remove it.

Makefile
masayag: I'd prefer to have this deterministic and explicit: if someone isn't up-do-date on his env, the result will look different than what is expected. Also, by having it explicit, one can easily rebuild a specific version.

pkliczewski: ok, make sense

Makefile
masayag: @jakub-dzon update-codegen.sh is broken. let's fix it separately and then include it.

```
→ ./hack/update-codegen.sh 
bash: ../code-generator/generate-groups.sh: No such file or directory
```

Makefile
masayag: when running `make docker-build` it will use $USER, in order apply a different user run with:
`QUAY_USER=other_user make build-docker push-docker`

pkliczewski (approved):

jakub-dzon (approved):

PR #27, opened 2020-03-30
Add tests

machacekondra:
Signed-off-by: Ondra Machacek <omachace@redhat.com>

PR #152, opened 2020-04-27
Basic e2e tests framework

jakub-dzon:
framework package has been based on https://github.com/kubevirt/containerized-data-importer/blob/master/tests/framework/framework.go

```release-note
NONE
```


tests/basic_vm_import_test.go
```diff
@@ -0,0 +1,66 @@
...
+	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
+	"github.com/kubevirt/vm-import-operator/tests/framework"
+	. "github.com/kubevirt/vm-import-operator/tests/matchers"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TODO: for now running on machine with access to oVirt hosting cirrosvm VM is required.
```
pkliczewski: Why not to use locally ovirt [mocks](https://github.com/kubevirt/vm-import-operator/pull/27/files#diff-a0068931f00eb90657e4a876bc012afbR1) that Ondra created some time ago.

tests/basic_vm_import_test.go
```diff
@@ -0,0 +1,66 @@
...
+			Fail("Cannot create secret: " + err.Error())
+		}
+		secret = s
+	})
+
+	Context(" without resource mapping", func() {
+		It("should create stopped VM", func() {
+			ovirtSecretName := secret.Name
+			// ID of a VM existing in oVirt described by the secret.
+			vmID := "c1a784d6-4747-4c26-8b47-d3d6766436e1"
```
pkliczewski: If we decide to use Ondra's mocks this `ID` is OK but with real ovirt we would need to fetch this value.

tests/basic_vm_import_test.go
jakub-dzon: The automation env scripts are not merged yet. When I'm sure I can use them and mocks, I will use them. 
It's WIP and is only supposed to show general direction of the work.

tests/basic_vm_import_test.go
jakub-dzon: The desired state is having a mock that we can instruct to return specific objects.

pkliczewski (approved):

machacekondra (approved):

PR #159, opened 2020-04-28
Fix path to manifests in tests

machacekondra:
This PR removes `cluster-up` and `cluster-sync` directories and add just single `cluster` directory, which hold all bash scripts used to setup/teardow/sync cluster.


pkliczewski (approved):

cluster-sync/sync.sh
```diff
@@ -26,9 +26,10 @@ if [ "${KUBEVIRT_PROVIDER}" != "external" ]; then
 fi
 
 # TODO: fetch it dynamically
-export OPERATOR_VERSION=0.0.1
+export OPERATOR_VERSION=v0.0.1
```
masayag: We can update `hack/prepare-release.sh` to replace the versions to be explicit same as it does for README file - so we want have to manually manage this file.

cluster-sync/sync.sh
```diff
@@ -26,9 +26,10 @@ if [ "${KUBEVIRT_PROVIDER}" != "external" ]; then
...
 
-_kubectl apply -f manifests/vm-import-operator/${OPERATOR_VERSION}/v2v_v1alpha1_resourcemapping_crd.yaml
-_kubectl apply -f manifests/vm-import-operator/${OPERATOR_VERSION}/v2v_v1alpha1_virtualmachineimport_crd.yaml
-_kubectl apply -f manifests/vm-import-operator/${OPERATOR_VERSION}/config_map.yaml
-_kubectl apply -f manifests/vm-import-operator/${OPERATOR_VERSION}/operator.yaml
+_kubectl apply -f https://github.com/kubevirt/vm-import-operator/releases/download/${OPERATOR_VERSION}/namespace.yaml
```
jakub-dzon: We need to deploy the manifests into the same namespace as kubevirt and for that the files need to be processed before being applied.

With local files we can run `TARGET_NAMESPACE=kubevirt make gen-manifests` and we're ready to test.

jakub-dzon (changes requested):

cluster-sync/sync.sh
machacekondra: Right, thanks. I've just change it so we generate it for each test, so we call:
`VERSION=${OPERATOR_VERSION} TARGET_NAMESPACE=${OPERATOR_NAMESPACE} make gen-manifests`

cluster-sync/sync.sh
machacekondra: Right, thanks. I've changed it so we use `make gen-manifests`, and always latest operator from code, not latest public version.

cluster-sync/sync.sh
```diff
@@ -25,10 +25,19 @@ if [ "${KUBEVIRT_PROVIDER}" != "external" ]; then
...
-
+# Build the operator:
+export RELEASE_VERSION=v0.15.2
+export GO111MODULE=on
+export OPERATOR_NAMESPACE=kubevirt
+export OPERATOR_VERSION=latest
+VERSION=${OPERATOR_VERSION} TARGET_NAMESPACE=${OPERATOR_NAMESPACE} make gen-manifests
+curl -LO https://github.com/operator-framework/operator-sdk/releases/download/${RELEASE_VERSION}/operator-sdk-${RELEASE_VERSION}-x86_64-linux-gnu
+chmod +x operator-sdk-${RELEASE_VERSION}-x86_64-linux-gnu && sudo mkdir -p /usr/local/bin/ && sudo cp operator-sdk-${RELEASE_VERSION}-x86_64-linux-gnu /usr/local/bin/operator-sdk && rm operator-sdk-${RELEASE_VERSION}-x86_64-linux-gnu
+operator-sdk --verbose build quay.io/kubevirt/vm-import-opreator:${OPERATOR_VERSION}
```
masayag: I think we need to replace this with a verb in Makefile since it already fetches the operator-sdk and capable to prepare its binary.
It will reduce the need to maintain the version of the operator-sdk outside of go.mod
https://github.com/kubevirt/cluster-network-addons-operator/blob/master/Makefile#L109

I'd also suggest to separate cluster-up / cluster-sync and operator-install into a separate step/file, i.e.
https://github.com/kubevirt/cluster-network-addons-operator/blob/master/cluster/operator-install.sh

So we'll get a clear view of the building blocks for test env, and all can be driven from the Makefile

machacekondra: That's nice repo with nice examples, thanks for pointing it out, I wasn't aware of that.

jakub-dzon (approved):

masayag (approved):

PR #170, opened 2020-04-30
Networked VM e2e tests

jakub-dzon:
```release-note
NONE
```

### DIFF
--- a/cluster-sync/imageio/imageio.yaml
+++ b/cluster-sync/imageio/imageio.yaml
@@ -22,7 +22,8 @@ spec:
         ports:
         - containerPort: 12345
       - name: fakeovirt
-        image: machacekondra/fakeovirt:test
+        # update back to original image when changes are merged
+        image: quay.io/jdzon/fakeovirt:latest
         imagePullPolicy: IfNotPresent
         ports:
         - containerPort: 12346
--- a/cluster-sync/imageio/imageio.yaml
+++ b/cluster-sync/imageio/imageio.yaml
@@ -22,8 +22,7 @@ spec:
         ports:
         - containerPort: 12345
       - name: fakeovirt
-        # update back to original image when changes are merged
-        image: quay.io/jdzon/fakeovirt:latest
+        image: machacekondra/fakeovirt:test
         imagePullPolicy: IfNotPresent
         ports:
         - containerPort: 12346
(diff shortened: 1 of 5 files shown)


cluster-sync/imageio/imageio.yaml
```diff
@@ -22,7 +22,8 @@ spec:
         ports:
         - containerPort: 12345
       - name: fakeovirt
-        image: machacekondra/fakeovirt:test
+        # update back to original image when changes are merged
```
pkliczewski: How do we track this change?

cluster-sync/imageio/imageio.yaml
```diff
@@ -22,7 +22,8 @@ spec:
...
+        image: quay.io/jdzon/fakeovirt:latest
```
machacekondra: I've updated the repo with your changes, thanks.

cluster-sync/imageio/imageio.yaml
jakub-dzon: Ondra has pushed the new image - I will update it in a moment.

cluster-sync/imageio/imageio.yaml
jakub-dzon: Thanks!

cluster-sync/imageio/imageio.yaml
jakub-dzon: Done.

tests/basic_net_vm_import_test.go
```diff
@@ -0,0 +1,101 @@
...
+var _ = Describe("Networked VM import ", func() {
+	var (
+		f         = framework.NewFrameworkOrDie("networked-vm-import")
+		secret    corev1.Secret
+		namespace string
+		test      = networkedVmImportTest{}
+	)
+
+	BeforeEach(func() {
+		namespace = f.Namespace.Name
```
masayag, line 31: Is the namespace on which the operator is deployed is used also for creating the vm-import resources and the VM ?
If this is the case, perhaps we'd like to make sure they are different, so there aren't any assumptions regarding permissions ?

jakub-dzon: No, it's not. 
`framework.Framework` creates namespace with generated name and prefix given in its constructor:
https://github.com/kubevirt/vm-import-operator/blob/b21b0d4961f4789b51292ea715f1e7186ac96f7e/tests/framework/framework.go#L159

pkliczewski (approved):

masayag (approved):

machacekondra (approved):

PR #171, opened 2020-05-01
Use vNIC Profile as source network mappings for oVirt

### LINKED ISSUE #111: Use vNic Profile in in source network mappings
Instead using networks within networkMappings, we need to have the vnic profile as identifier of that network, since it is possible that two vnic profiles of the same network will be mapped to two different network attachment definition resources on kubevirt.

masayag:
Source Network from oVirt can be mapped via vNIC Profiles to different
network attachment definition entities on KubeVirt.
Hence the semantic of the network mapping under 'ovirt' source of the
resource mappings is changed to support it.

- _ovirt.networkMappings.source.name_ expects the format of _'network-name/vnic-profile-name'_ 
- _ovirt.networkMappings.source.name_ expects the value of the vNIC Profile ID.

Fixes #111

### DIFF
--- a/pkg/providers/ovirt/validation/validators/network-mapping-validator.go
+++ b/pkg/providers/ovirt/validation/validators/network-mapping-validator.go
@@ -2,6 +2,7 @@ package validators
 
 import (
 	"fmt"
+	"strings"
 
 	netv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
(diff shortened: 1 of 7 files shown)


pkg/providers/ovirt/validation/validators/network-mapping-validator.go
```diff
@@ -89,6 +98,24 @@ func (v *NetworkMappingValidator) hasAtLeastOneWithVNicProfile(nics []*ovirtsdk.
 	return false
 }
 
+func (v *NetworkMappingValidator) validateSourceNetworkFormat(mapByName map[string]v2vv1alpha1.ResourceMappingItem) (ValidationFailure, bool) {
+	invalidNames := make([]string, 0)
+	for k := range mapByName {
+		if !strings.Contains(k, "/") {
```
pkliczewski, line 104: Is it possible that either network name or vnicprofile name could contain `/`?

masayag: it seems that both might contain any set of characters - the restriction for network/vnic name was removed https://gerrit.ovirt.org/#/c/79459/ 
it even seems legal for vnic profile name or network name to contain only `/////` 

However, that doesn't break the semantic of this field. If there is no `/` - it means it doesn't contain both elements. If it contains at least `/` and it doesn't match any of vm nics - it should fail.

The alternative is suggesting a new element to identify the network name, so the element will look like as:
```
networkMappings:
   source:
      vnicProfileName: profile1
      networkName: red
      vnicProfileID: xxx-xxx-xxx
```
I'd prefer staying with current format of a common `ResourceMappingItem` to identify sources by ID or name and not suggest element for each source.

pkliczewski (approved):

PR #172, opened 2020-05-03
Use Vnic Profile as source mapping

### LINKED ISSUE #111: Use vNic Profile in in source network mappings

masayag:
This PR completes missing pieces of https://github.com/kubevirt/vm-import-operator/pull/171
While #171 deals with validations, this PR deals with mappings.

Closes #111

### DIFF
--- a/pkg/providers/ovirt/mapper/mapper.go
+++ b/pkg/providers/ovirt/mapper/mapper.go
@@ -7,6 +7,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 
 	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
+	outils "github.com/kubevirt/vm-import-operator/pkg/providers/ovirt/utils"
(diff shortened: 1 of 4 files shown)


pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -285,13 +286,14 @@ func (o *OvirtMapper) mapNics(networkToType map[string]string) []kubevirtv1.Inte
 	var kubevirtNics []kubevirtv1.Interface
 	nics, _ := o.vm.Nics()
 	for _, nic := range nics.Slice() {
-		// This network interface don't have any network specified.
+		// This network interface doesn't have any network specified.
```
pkliczewski: Shouldn't this be ... doesn't have any vnic profile specified?

pkliczewski (approved):

machacekondra (approved):

jakub-dzon (approved):

PR #175, opened 2020-05-04
Adding tablet input device when VNC is enabled

### LINKED ISSUE #160: VNC not working properly
VNC cursor is buggy in web console UI due to missing tablet device. 
BZ bug: https://bugzilla.redhat.com/show_bug.cgi?id=1827705

The templates itself are not perfect (yet) as they (together with kubevirt) configure VNC by default to all VMs.

We do not have to rely on common templates for this feature as we can detect if oVirt VM has VNC configured or not. And set autoattachGraphicsDevice to true|false accordingly. Beware that this attribute defaults to true if missing (https://kubevirt.io/api-reference/master/definitions.html#_v1_devices)

input device should be present in `devices.inputs` when setting autoattachGraphicsDevice to true for VNC to work properly:

for windows:

```
- bus: usb
  name: tablet
  type: tablet
```

and for linux

```
- bus: virtio
  name: tablet
  type: tablet
```

this should be applied to all templates or empty VM.

jakub-dzon:
Fixes #160 

```release-note
NONE
```

### DIFF
--- a/pkg/providers/ovirt/mapper/mapper.go
+++ b/pkg/providers/ovirt/mapper/mapper.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strings"
 
+	oos "github.com/kubevirt/vm-import-operator/pkg/providers/ovirt/os"
+
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 
 	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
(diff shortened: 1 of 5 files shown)


pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -610,6 +619,19 @@ func (o *OvirtMapper) mapGraphicalConsoles() *kubevirtv1.Devices {
 	return devices
 }
 
+func (o *OvirtMapper) prepareTabletInputDevice(os string) kubevirtv1.Input {
+	tablet := kubevirtv1.Input{
+		Type: "tablet",
+		Name: "tablet",
+	}
+	if strings.Contains(os, "win") {
```
pkliczewski: There could be case sensitivity issue here. Wouldn't be better to use https://golang.org/pkg/strings/#EqualFold?

jakub-dzon: :+1: Thank you!

pkliczewski (approved):

masayag (approved):

PR #184, opened 2020-05-06
Add events emitting

### LINKED ISSUE #143: Add support for events of VirtualMachineImport objects
It would be very useful for debugging (and visibility in the UI) to add events support. These should at least include important changes in Conditions and errors.

machacekondra:
This PR adds event emitting for the VM import flow.
Fixes: https://github.com/kubevirt/vm-import-operator/issues/143

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
 	kubevirtv1 "kubevirt.io/client-go/api/v1"
 	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -46,6 +47,17 @@ const (
 	progressDone                      = "100"
 	progressForCopyDisk               = 40
 	requeueAfterValidationFailureTime = 5 * time.Second
+
+	// EventImportSucceeded is emitted
+	EventImportSucceeded = "ImportSucceeded"
+	// EventImportBlocked is emitted
+	EventImportBlocked = "ImportBlocked"
+	// EventVMStartFailed is emitted
+	EventVMStartFailed = "VMStartFailed"
+	// EventVMCreationFailed is emitted
+	EventVMCreationFailed = "VMCreationFailed"
+	// EventDVCreationFailed is emitted
+	EventDVCreationFailed = "DVCreationFailed"
 )
 
 var (
@@ -81,6 +93,7 @@ func newReconciler(mgr manager.Manager, kvConfigProvider config.KubeVirtConfigPr
 		ownerreferencesmgr:     ownerreferencesmgr,
 		factory:                factory,
 		kvConfigProvider:       kvConfigProvider,
+		recorder:               mgr.GetEventRecorderFor("virtualmachineimport-controller"),
 	}
 }
 
@@ -142,6 +155,7 @@ type ReconcileVirtualMachineImport struct {
 	ownerreferencesmgr     ownerreferences.OwnerReferenceManager
 	factory                pclient.Factory
 	kvConfigProvider       config.KubeVirtConfigProvider
+	recorder               record.EventRecorder
 }
 
 // Reconcile reads that state of the cluster for a VirtualMachineImport object and makes changes based on the state read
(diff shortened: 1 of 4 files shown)


pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -255,6 +269,9 @@ func (r *ReconcileVirtualMachineImport) importDisks(provider provider.Provider,
 					if err := r.afterSuccess(vmName, vmiName, provider); err != nil {
 						return err
 					}
+
+					// Emit event vm is successfully imported
+					r.recorder.Event(instance, corev1.EventTypeNormal, EventImportSucceeded, "Virtual Machine import successful")
```
pkliczewski: it would be good to add vm name so the user would know which vm was imported.
I would rephrase it to "Virtual Machine %s was imported successfully"

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -314,8 +331,12 @@ func (r *ReconcileVirtualMachineImport) createVM(provider provider.Provider, ins
 	// Create kubevirt VM from source VM:
 	reqLogger.Info("Creating a new VM", "VM.Namespace", vmSpec.Namespace, "VM.Name", vmSpec.Name)
 	if err = r.client.Create(context.TODO(), vmSpec); err != nil && !errors.IsAlreadyExists(err) {
+		message := fmt.Sprintf("Error while creating virtual machine: %s", err)
```
pkliczewski: Can we add vm name here as well?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -362,10 +383,15 @@ func (r *ReconcileVirtualMachineImport) startVM(provider provider.Provider, inst
 				return err
 			}
 			if err = r.updateToRunning(vmName); err != nil {
+				// Emit event vm failed to start:
+				r.recorder.Eventf(instance, corev1.EventTypeWarning, EventVMStartFailed, "Virtual Machine failed to start: %s", err)
```
pkliczewski: Which vm failed to start?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -445,6 +471,7 @@ func (r *ReconcileVirtualMachineImport) createDataVolumes(provider provider.Prov
 	// Fetch VM:
 	vmDef := &kubevirtv1.VirtualMachine{}
 	err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: vmName.Namespace, Name: vmName.Name}, vmDef)
+
```
pkliczewski, line 474: I think that this blank line is not needed.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -457,8 +484,12 @@ func (r *ReconcileVirtualMachineImport) createDataVolumes(provider provider.Prov
 
 		err = r.client.Create(context.TODO(), &dv)
 		if err != nil {
+			message := fmt.Sprintf("Data volume creation failed: %s", err)
```
pkliczewski: Would be good to add a bit of context about which datavolume for which vm.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -821,6 +853,10 @@ func (r *ReconcileVirtualMachineImport) validate(instance *v2vv1alpha1.VirtualMa
 		}
 		if valid, message := shouldFailWith(conditions); !valid {
 			logger.Info("Import blocked. " + message)
+
+			// Emit event vm import blocked:
+			r.recorder.Eventf(instance, corev1.EventTypeNormal, EventImportBlocked, "Virtual Machine import blocked: %s", message)
```
pkliczewski: please add vm name

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
masayag: pls use Namespace/Name for VM

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
jakub-dzon: If the VM is blocked, that event will be emitted on every reconciliation until the cause for the block is fixed; I think we might need some throttling around recording of that event.

machacekondra: Well, we can see following in the events of vm import:
`
  Normal  ImportBlocked  1s (x23 over 79s)  virtualmachineimport-controller  Virtual Machine import blocked: VM uses unsupported bios type: i440fx_sea_bios
`

and following in get events:
`
50s         Normal    ImportBlocked             virtualmachineimport/myvm                                    Virtual Machine import blocked: VM uses unsupported bios type: i440fx_sea_bios
`

I think it's fine, it's not flooded, you just see counter. Or I can check whether this event was already generated for this VM import and don't generate it if it was.

jakub-dzon: It does not look that bad. I wonder if in case of mass import, when there is some issue with, for example, resource mapping for tens or hundreds of VMs the event rate would be problematic. My gut tells me that for a good event processing service thousands of events per second should not be a problem.

pkliczewski (approved):

jakub-dzon (approved):

masayag (approved):

PR #186, opened 2020-05-06
Build operator by 'make docker-build'

masayag:


masayag:
The only discrepancy i see with it is that csv-generator is built from within the Dockerfile, and the operator is being built explicitly.

I could have just added the  operator-build to Dockerfile, but then it is being used by `operator-sdk build` which doesn't achieve the desired goal and will override the compiled version by the operator-sdk.

pkliczewski (approved):

jakub-dzon (approved):

PR #189, opened 2020-05-07
'latest' is not a semver and cannot be used with csv-generator

jakub-dzon:
"'latest" is not a semver and cannot be used with csv-generator:                                                                          
```
./hack/generate-manifests.sh
panic: No Major.Minor.Patch elements found
 
goroutine 1 [running]:
main.main()
        /home/jdzon/go/src/github.com/kubevirt/vm-import-operator/tools/csv-generator/csv-generator.go:38 +0x2a1
```
```release-note
NONE
```

### DIFF
--- a/cluster/operator-install.sh
+++ b/cluster/operator-install.sh
@@ -16,7 +16,7 @@
 
 set -ex
 
-export VERSION=latest
+export VERSION=v0.0.0
 
 ./cluster/kubectl.sh create -f _out/vm-import-operator/${VERSION}/operator.yaml
 ./cluster/kubectl.sh wait deploy/vm-import-operator -n kubevirt --for=condition=Available --timeout=600s
--- a/cluster/operator-push.sh
+++ b/cluster/operator-push.sh
@@ -16,7 +16,7 @@
 
 set -ex
 
-export VERSION=latest
+export VERSION=v0.0.0
 
 registry_port=$(./cluster/cli.sh ports registry | tr -d '\r')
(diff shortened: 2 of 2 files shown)


cluster/operator-install.sh
```diff
@@ -16,7 +16,7 @@
 
 set -ex
 
-export VERSION=latest
+export VERSION=v0.0.0-1
```
pkliczewski: Why not to use operator version -> v0.0.1?

jakub-dzon: I wanted to make sure that the version built for the tests is different than one that we would release.

pkliczewski: What about using v0.0.0 like in go.mod when we relay on latest?

jakub-dzon: :+1:

masayag:
In which step this error was detected ? what is a part of the e2e test run ?

jakub-dzon:
Yes, the error originated from  `make gen-manifests` https://github.com/kubevirt/vm-import-operator/blob/d35bc73e47002121aa185ed68a9d89e54ecb09d5/cluster/operator-push.sh#L32
called by `make cluster-sync` from
https://github.com/kubevirt/vm-import-operator/blob/d35bc73e47002121aa185ed68a9d89e54ecb09d5/automation/test.sh#L31

masayag (approved):

pkliczewski (approved):

PR #190, opened 2020-05-07
hco: add vmimport config cr

pkliczewski:
Add vm import config custom resource need by the HCO.

### DIFF
--- a/manifests/vm-import-operator/v0.0.1/v2v.kubevirt.io_vmimportconfigs_cr.yaml
+++ b/manifests/vm-import-operator/v0.0.1/v2v.kubevirt.io_vmimportconfigs_cr.yaml
@@ -1,0 +1,7 @@
+---
+apiVersion: v2v.kubevirt.io/v1alpha1
+kind: VMImportConfig
+metadata:
+  name: vm-import-operator-config
+spec:
+  imagePullPolicy: IfNotPresent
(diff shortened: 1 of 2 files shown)


manifests/vm-import-operator/v0.0.1/v2v.kubevirt.io_vmimportconfigs_cr.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: v2v.kubevirt.io/v1alpha1
+kind: VMImportConfig
+metadata:
+  name: vm-import-operator-config
```
masayag, line 5: since this is namespaced scope resource, does it make sense to add `namespace: kubevirt-hyperconverged` ?
or does the namespace should be decided by the context in which this resource will be installed ?
Should we add it as a template and generate its output to replace the namespace value ?

pkliczewski: I checked SSP and CDI CRs and there is no namespace.

masayag: so i think we need to make sure it is being added to each release by adding it as 
`v2v.kubevirt.io_vmimportconfigs_cr.yaml.in` under `templates/vm-import-operator/VERSION`

pkliczewski: ok, will add

masayag (approved):

PR #192, opened 2020-05-07
Install golang during up.sh

machacekondra:

### DIFF
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -29,6 +29,8 @@ if [[ "$KUBEVIRT_PROVIDER" =~ (ocp|okd)- ]]; then
     ./cluster/kubectl.sh delete ns openshift-cluster-samples-operator --wait=false
 fi
 
+ensure_golang
+
 install_cdi
 install_kubevirt
 install_imageio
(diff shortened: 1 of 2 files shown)


cluster/up.sh
```diff
@@ -29,6 +29,9 @@ if [[ "$KUBEVIRT_PROVIDER" =~ (ocp|okd)- ]]; then
     ./cluster/kubectl.sh delete ns openshift-cluster-samples-operator --wait=false
 fi
 
+# Install golang to run generate manifests
+sudo yum install -y golang
```
jakub-dzon: That will be run on the machine where the tests are executed and if the tests are run locally, it will always ask for the root password (and then try to install golang tools unnecessarily). Do we need that?

masayag: this is a valid question. if you run as a non-root/non-sudoer user - perhaps it is preferred to install go by downloading it
using wget & install it locally.

or is there an option to require the VM/OS to have go installed as a prerequisite ?

machacekondra: For now we support only rpm based OSes. So I've added an `if` which check if `golang` is already installed, and skip the sudo install part if yes.

cluster/up.sh
```diff
@@ -29,6 +29,8 @@ if [[ "$KUBEVIRT_PROVIDER" =~ (ocp|okd)- ]]; then
     ./cluster/kubectl.sh delete ns openshift-cluster-samples-operator --wait=false
 fi
 
+check_golang
```
masayag: ensure_golang will describe it more accurately

jakub-dzon (approved):

masayag (approved):

PR #193, opened 2020-05-07
Resource mapping tests

jakub-dzon:
```release-note
NONE
```

### DIFF
--- a/tests/basic_vm_import_test.go
+++ b/tests/basic_vm_import_test.go
@@ -14,23 +14,20 @@ import (
 	v1 "kubevirt.io/client-go/api/v1"
 )
 
-var (
-	// ID of a VM existing in oVirt described by the secret.
-	vmID = "123"
-
-	trueVar = true
-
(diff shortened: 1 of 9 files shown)


tests/basic_vm_import_test.go
```diff
@@ -14,23 +14,20 @@ import (
...
-	storageClass = "local"
-)
-
 type basicVmImportTest struct {
 	framework *fwk.Framework
 }
 
 var _ = Describe("Basic VM import ", func() {
+	var (
+		vmID = "123"
```
pkliczewski, line 23: What do you think about moving those constants somewhere from where we can reference them in many tests?
We could do it in different PR.

jakub-dzon: :+1: I thought about it as well, will do it in a separate PR.

pkliczewski (approved):

masayag (approved):

PR #198, opened 2020-05-10
Allow user to override OS Map

masayag:
In order to simplify deployment of the operator, and due to the fact
that OS map is part of the operator, the core OS Map is being embedded
to the operator's code.

In addition, in order to allow user to specify a custom OS map (to
extend or override core OS Map), the deployment of the operator
introduces two environment variables that will be used to identify the
user OS Map by name and namespace:
- OS_CONFIGMAP_NAME
- OS_CONFIGMAP_NAMESPACE

### DIFF
--- a/pkg/os/os-mapper-provider.go
+++ b/pkg/os/os-mapper-provider.go
@@ -2,18 +2,27 @@ package os
 
 import (
 	"context"
+	"fmt"
+	"os"
 
 	yaml "gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
-	clientutil "kubevirt.io/client-go/util"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	osConfigMapName    = "vmimport-os-mapper"
-	guestOsToCommonKey = "guestos2common"
-	osInfoToCommonKey  = "osinfo2common"
+	// GuestOsToCommonKey represents the element key in OS map of guest OS to common template mapping
+	GuestOsToCommonKey = "guestos2common"
+
+	// OsInfoToCommonKey represents the element key in OS map of OS type to common template mapping
+	OsInfoToCommonKey = "osinfo2common"
+
+	// OsConfigMapName represents the environment variable name that holds the OS config map name
+	OsConfigMapName = "OS_CONFIGMAP_NAME"
+
+	// OsConfigMapNamespace represents the environment variable name that holds the OS config map namespace
+	OsConfigMapNamespace = "OS_CONFIGMAP_NAMESPACE"
 )
 
 // OSMaps is responsible for getting the operating systems maps which contain mapping of GuestOS to common templates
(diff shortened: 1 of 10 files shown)


pkliczewski (approved):

pkg/os/os-mapper-provider.go
```diff
@@ -36,25 +45,104 @@ func NewOSMapProvider(client client.Client) *OSMaps {
 
 // GetOSMaps retrieve the OS mapping config map
 func (o *OSMaps) GetOSMaps() (map[string]string, map[string]string, error) {
+	guestOsToCommon := initGuestOsToCommon()
+	osInfoToCommon := initOsInfoToCommon()
+	err := o.updateOsMapsByUserMaps(guestOsToCommon, osInfoToCommon)
+	return guestOsToCommon, osInfoToCommon, err
+}
+
+func (o *OSMaps) updateOsMapsByUserMaps(guestOsToCommon map[string]string, osInfoToCommon map[string]string) error {
```
jakub-dzon, line 54: What do you think about renaming it to `updateOsMapsWithUserMaps`?

pkg/os/os-mapper-provider.go
```diff
@@ -36,25 +45,104 @@ func NewOSMapProvider(client client.Client) *OSMaps {
...
+			"Failed to read user OS config-map [%s/%s] due to: [%v]",
+			osConfigMapNamespace,
+			osConfigMapName,
+			err,
+		)
 	}
-	guestOsToCommon := make(map[string]string)
-	err = yaml.Unmarshal([]byte(osConfigMap.Data[guestOsToCommonKey]), &guestOsToCommon)
+
+	err = yaml.Unmarshal([]byte(osConfigMap.Data[GuestOsToCommonKey]), &guestOsToCommon)
```
jakub-dzon, line 76: Does `yaml.Unmarshall` merge parsed data into the map passed in parameter or replace its content completely?

machacekondra (approved):

pkg/os/os-mapper-provider_test.go
```diff
@@ -0,0 +1,164 @@
...
+				obj.(*corev1.ConfigMap).Data = make(map[string]string)
+				obj.(*corev1.ConfigMap).Data[osmap.GuestOsToCommonKey] = `"Fedora": "myfedora"`
+				obj.(*corev1.ConfigMap).Data[osmap.OsInfoToCommonKey] = `"rhel_6_9_plus_ppc64": "myos"`
+				return nil
+			}
+
+			guestOsToCommon, osInfoToCommon, err := osMapper.GetOSMaps()
+
+			Expect(err).NotTo(HaveOccurred())
+			// core values should be loaded prior to loading user OS map data
```
jakub-dzon, line 87: Depending of the intended behaviour of I would suggest following to make the API behaviour explicitly described:
- if the map contents are merged - add check that one of the core mappings still works;
- otherwise check that one of the core ones is not there;

jakub-dzon (changes requested):

pkg/os/os-mapper-provider.go
masayag: yes, it merged the values. I'll add checks to the test that when a value is overridden, the total sum of elements is greater than the provided map size.

PR #203, opened 2020-05-11
hco: vmimport operator

pkliczewski:
We add new operator responsible for deployment of vm-import-operator based on VMImportConfig object used by HCO.


pkg/operator/resources/operator/operator.go
```diff
@@ -47,19 +49,45 @@ type csvStrategySpec struct {
...
 		ObjectMeta: metav1.ObjectMeta{
-			Name:   name,
+			Name:   "vm-import-operator",
 			Labels: map[string]string{},
 		},
 	}
 }
 
+// CreateRoleBinging returns role binging for vm-import-operator
+func CreateRoleBinging(namespace string) *rbacv1.RoleBinding {
```
jakub-dzon: `CreateRoleBinding`

pkliczewski: Thanks!

Makefile
```diff
@@ -2,19 +2,18 @@ all: fmt check
 
 # Always keep the last released version here
 VERSION_REPLACES ?= v0.0.0
-VERSION ?= v0.0.1
+VERSION ?= v0.0.2
```
masayag: just repeating my comment here as well - should be reverted back to v0.0.1

Makefile
```diff
@@ -2,19 +2,18 @@ all: fmt check
...
 TARGET_NAMESPACE ?= kubevirt-hyperconverged
-export TARGET_NAMESPACE := $(TARGET_NAMESPACE)
-
 DEPLOY_DIR ?= manifests
 
 # Image registry variables
 QUAY_USER ?= $(USER)
 IMAGE_REGISTRY ?= quay.io/$(QUAY_USER)
-IMAGE_TAG ?= latest
+IMAGE_TAG ?= $(VERSION)
```
masayag: Would that require a change in one of the e2e tests ? IIRC, there was an assumption that the latest build should be used. @machacekondra do you recall in which script the image is being built for testing ?

README.md
```diff
@@ -191,6 +191,22 @@ Connect to the debug session, i.e. if using vscode, create launch.json as:
 }
 ```
 
+# Running tests
```
masayag: In a latest stage, I'd pull the developer aspects into a separate document (perhaps move it to docs/contributing.md or docs/developers.md) with other aspects that concerns development & testing

pkg/apis/v2v/v1alpha1/vmimportconfig_types.go
```diff
@@ -37,8 +40,33 @@ type VMImportConfigStatus struct {
...
 	ObservedVersion string `json:"observedVersion,omitempty" optional:"true"`
+	// +optional
+	Phase VMImportPhase `json:"phase,omitempty"`
 }
 
+const (
+	// PhaseDeploying signals that the resources are being deployed
+	PhaseDeploying VMImportPhase = "Deploying"
+
+	// PhaseDeployed signals that the resources are successflly deployed
```
masayag: s/successflly/successfully

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+	r, err := newReconciler(mgr)
+	if err != nil {
+		return err
+	}
+	return r.add(mgr)
+}
+
+// OperatorArgs contains the required parameters to generate all namespaced resources
+type OperatorArgs struct {
+	OperatorVersion        string `required:"true" split_words:"true"`
```
masayag, line 63: What are the `required:"true" split_words:"true" used for ?

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+	targetVersion = strings.TrimPrefix(targetVersion, "v")
+	currentVersion = strings.TrimPrefix(currentVersion, "v")
+
+	// our default position is that this is an update.
+	// So if the target and current version do not
+	// adhere to the semver spec, we assume by default the
+	// update path is the correct path.
+	shouldTakeUpdatePath := true
+	target, err := semver.Make(targetVersion)
+	if err == nil {
```
masayag: i'd reduce the level of nesting by early quit (in this case `shouldTakeUpdatePath` can be spared)

```
if err != nil {
    return true, nil
}

current, err := semver.Make(currentVersion)
if err == nil {
    return true, nil
}

if target.Compare(current) < 0 {
        err := fmt.Errorf("operator downgraded, will not reconcile")
	logger.Error(err, "", "current", current, "target", target)
	return false, err
}

if target.Compare(current) == 0 {
	return false, nil
}

return true, nil
```

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+
+func (r *ReconcileVMImportConfig) checkUpgrade(logger logr.Logger, cr *vmimportv1alpha1.VMImportConfig) error {
+	// should maybe put this in separate function
+	if cr.Status.OperatorVersion != r.operatorArgs.OperatorVersion {
+		cr.Status.OperatorVersion = r.operatorArgs.OperatorVersion
+		cr.Status.TargetVersion = r.operatorArgs.OperatorVersion
+		if err := r.crUpdate(cr.Status.Phase, cr); err != nil {
+			return err
+		}
+	}
```
masayag, line 229: this part updates the CR, while the method name is `checkUpgrade` which semantically sounds like it shouldn't change the state of the resource. 
Consider changing it to `updateResourceForUpgrade`

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+
+	isUpgrade, err := shouldTakeUpdatePath(logger, r.operatorArgs.OperatorVersion, cr.Status.ObservedVersion)
+	if err != nil {
+		return err
+	}
+
+	if isUpgrade && cr.Status.Phase != vmimportv1alpha1.PhaseUpgrading {
+		logger.Info("Observed version is not target version. Begin upgrade", "Observed version ", cr.Status.ObservedVersion, "TargetVersion", r.operatorArgs.OperatorVersion)
+		MarkCrUpgradeHealingDegraded(cr, "UpgradeStarted", fmt.Sprintf("Started upgrade to version %s", r.operatorArgs.OperatorVersion))
```
masayag: "UpgradeStarted" can be replaced with a type definition VmConfigUpgradeReason (or other name of choice) to be added to the vmconfig_types.go file, where all of the reasons can be added and documented

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+		if err := r.crUpdate(vmimportv1alpha1.PhaseUpgrading, cr); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// MarkCrUpgradeHealingDegraded marks the passed CR as upgrading and degraded.
+func MarkCrUpgradeHealingDegraded(cr *vmimportv1alpha1.VMImportConfig, reason, message string) {
```
masayag: the visibility can be reduced if there is no intent to export this method

machacekondra (approved):

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+
+	return resources, nil
+}
+
+func createControllerResources(args *OperatorArgs) []runtime.Object {
+	return []runtime.Object{
+		resources.CreateServiceAccount(args.Namespace),
+		resources.CreateControllerRole(args.Namespace),
+		resources.CreateControllerRoleBinding(args.Namespace),
+		resources.CreateControllerDeployment("vm-import-controller", args.Namespace, args.ControllerImage, args.PullPolicy, int32(1)),
```
masayag: the first argument "vm-import-controller" seems redundant here - it should be encapsulated in resources.CreateControllerDeployment() since we don't expect user to set it with a different name

also, within `CreateControllerDeployment` we use the constant `controllerName` for other values of the operator.

which means that 614 https://github.com/kubevirt/vm-import-operator/pull/203/files#diff-c5c8aa376175139b13b0583b01ad56c6R614 should be changed to `vm-import-controller` (or the opposite - the first argument to this method should be named vm-import-deployment)

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+				shouldTakeUpdatePath = false
+			}
+		}
+	}
+
+	return shouldTakeUpdatePath, nil
+}
+
+func (r *ReconcileVMImportConfig) checkUpgrade(logger logr.Logger, cr *vmimportv1alpha1.VMImportConfig) error {
+	// should maybe put this in separate function
```
masayag: comment can be removed

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+			}
+
+			currentRuntimeObj = desiredRuntimeObj.DeepCopyObject()
+			logger.Info("Resource creating",
+				"namespace", desiredMetaObj.GetNamespace(),
+				"name", desiredMetaObj.GetName(),
+				"type", fmt.Sprintf("%T", desiredMetaObj))
+
+			if err = r.client.Create(context.TODO(), currentRuntimeObj); err != nil {
+				logger.Error(err, "")
```
masayag, line 307: will the message be detailed enough in case of an error ?

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+				allErrors = append(allErrors, err)
+				continue
+			}
+
+			logger.Info("Resource created",
+				"namespace", desiredMetaObj.GetNamespace(),
+				"name", desiredMetaObj.GetName(),
+				"type", fmt.Sprintf("%T", desiredMetaObj))
```
masayag: will it be possible to extract this to `createResourceForController` method ?
mainly since this method is pretty long

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+					"namespace", desiredMetaObj.GetNamespace(),
+					"name", desiredMetaObj.GetName(),
+					"type", fmt.Sprintf("%T", desiredMetaObj))
+			} else {
+				logger.V(3).Info("Resource unchanged",
+					"namespace", desiredMetaObj.GetNamespace(),
+					"name", desiredMetaObj.GetName(),
+					"type", fmt.Sprintf("%T", desiredMetaObj))
+			}
+		}
```
masayag: same here - will it be possible to extract it to `updateResourceForController` ?

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+	}
+	cr.Status.ObservedVersion = version
+	cr.Status.OperatorVersion = version
+	cr.Status.TargetVersion = version
+	return r.crUpdate(phase, cr)
+}
+
+func (r *ReconcileVMImportConfig) isMutable(obj runtime.Object) bool {
+	switch obj.(type) {
+	case *corev1.ConfigMap, *corev1.Secret, *rbacv1.RoleBinding, *rbacv1.Role:
```
masayag, line 856: no need to include here ClusterRole and ClusterRoleBinding ?

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+			currentRuntimeObj = desiredRuntimeObj.DeepCopyObject()
+			logger.Info("Resource creating",
+				"namespace", desiredMetaObj.GetNamespace(),
+				"name", desiredMetaObj.GetName(),
+				"type", fmt.Sprintf("%T", desiredMetaObj))
+
+			if err = r.client.Create(context.TODO(), currentRuntimeObj); err != nil {
+				logger.Error(err, "")
+				allErrors = append(allErrors, err)
+				continue
```
masayag, line 309: Why a failure to create a required resource shouldn't fail the rest of the process ?
Same goes for an error to update a resource. Wouldn't it leave the operator is semi-updated state ?

pkg/operator/controller/controller.go
```diff
@@ -0,0 +1,864 @@
...
+				logger.V(3).Info("Resource unchanged",
+					"namespace", desiredMetaObj.GetNamespace(),
+					"name", desiredMetaObj.GetName(),
+					"type", fmt.Sprintf("%T", desiredMetaObj))
+			}
+		}
+	}
+
+	if len(allErrors) > 0 {
+		return reconcile.Result{}, fmt.Errorf("reconcile encountered %d errors", len(allErrors))
```
masayag, line 347: no need to update the VMImportConfig status with the errors ?

pkg/operator/controller/controller_suite_test.go
```diff
@@ -0,0 +1,83 @@
...
+func TestOperator(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Operator Resource Suite")
+}
+
+var testenv *envtest.Environment
+var cfg *rest.Config
+var clientset *kubernetes.Clientset
+
+var crd = &extv1beta1.CustomResourceDefinition{
```
masayag: why not reuse CreateVMImportConfig() from pkg/operator/resources/operator/operator.go ?

pkg/operator/resources/operator/operator.go
```diff
@@ -47,20 +53,49 @@ type csvStrategySpec struct {
 	Deployments        []csvDeployments `json:"deployments"`
 }
 
-func createRole(name string) *rbacv1.Role {
+// CreateControllerRole returns role for vm-controller-operator
+func CreateControllerRole(namespace string) *rbacv1.Role {
```
masayag, line 57: the kind of the created role resource says `ClusterRole` but the type is `rbacv1.Role`
seems like a mismatch, so if it is namespaced-scope, it shouldn't be a ClusterRole and the kind should be updated.

pkg/operator/resources/operator/operator.go
```diff
@@ -47,20 +53,49 @@ type csvStrategySpec struct {
...
+			Labels:    map[string]string{},
+			Namespace: namespace,
+		},
+		Rules: getPolicyRules(),
+	}
+}
+
+// CreateControllerRoleBinding returns role binging for vm-import-operator
+func CreateControllerRoleBinding(namespace string) *rbacv1.RoleBinding {
+	return &rbacv1.RoleBinding{
```
masayag, line 74: same here - rbacv1.RoleBinding doesn't match kind ClusterRoleBinding

pkg/operator/resources/operator/operator.go
```diff
@@ -976,23 +1050,126 @@ the disk alias on ovirt DiskMappings is respected only when provided in context
 	}
 }
 
+func createOperatorDeployment(operatorVersion, namespace, deployClusterResources, operatorImage, controllerImage, pullPolicy string) *appsv1.Deployment {
+	deployment := CreateOperatorDeployment("vm-import-operator", namespace, "name", "vm-import-operator", serviceAccountName, int32(1))
```
masayag: the hard-coded values can be replaced with the constant `operatorName`

pkg/operator/resources/operator/operator.go
```diff
@@ -976,23 +1050,126 @@ the disk alias on ovirt DiskMappings is respected only when provided in context
...
+	container := CreateContainer("vm-import-operator", operatorImage, corev1.PullPolicy(pullPolicy))
```
masayag: same here

pkg/operator/resources/operator/operator.go
masayag: "vm-import-operator" can be replaced with constant operatorName

pkg/operator/resources/operator/operator.go
masayag: same here

masayag (changes requested):

Makefile
pkliczewski: I updated it in newer version. Thanks for pointing out.

Makefile
pkliczewski: Already updated. I wanted to test it with specific version.

README.md
pkliczewski: Good idea.

pkg/apis/v2v/v1alpha1/vmimportconfig_types.go
pkliczewski: Ack

pkg/operator/controller/controller.go
pkliczewski: Not sure, that was borrowed. I will remove.

pkg/operator/controller/controller.go
pkliczewski: Great, thanks will apply.

pkg/operator/controller/controller.go
pkliczewski: Ack

pkg/operator/controller/controller.go
pkliczewski: Good idea.

pkg/operator/controller/controller.go
pkliczewski: No need to export it, will reduce

pkg/operator/controller/controller.go
pkliczewski: I am using it for testing where I fake different object to make sure update works correctly.

pkg/operator/controller/controller.go
pkliczewski: :+1:

pkg/operator/controller/controller.go
pkliczewski: yes, there will be stacktrace with error message. It was useful when I debugged it.

pkg/operator/controller/controller.go
pkliczewski: Ack

pkg/operator/controller/controller.go
pkliczewski: Ack

pkg/operator/controller/controller.go
pkliczewski: I added them here based on my findings. Without that mutation code fails.

pkg/operator/controller/controller.go
pkliczewski: We have a concept of degraded from which we can recover in next reconcile. That is why we collect errors.

pkg/operator/controller/controller.go
pkliczewski: we could recover from it.

pkg/operator/controller/controller_suite_test.go
pkliczewski: Good idea.

pkg/operator/resources/operator/operator.go
pkliczewski: you are right, will fix.

pkg/operator/resources/operator/operator.go
pkliczewski: ack

pkg/operator/resources/operator/operator.go
pkliczewski: ack

pkg/operator/resources/operator/operator.go
pkliczewski: ack

pkg/operator/resources/operator/operator.go
pkliczewski: ack

pkg/operator/resources/operator/operator.go
pkliczewski: ack

pkg/operator/controller/controller.go
masayag: not sure you need to remove it, I was interested to know what it is used for:

 https://gitlab.com/gitlab-org/gitaly/commit/e32c0ad734100057234a2f9e2b255fc8ac3162d5#59d63d5e5529d0e3889addc3eed9bf040a85b8a9_0_98

pkg/operator/controller/controller.go
masayag: In that case, there will be a mismatch since the deployment is being created with name `vm-import-controller` but in method `isControllerDeployment` the deployment name is compared to `vm-import-deployment`
```
func isControllerDeployment(d *appsv1.Deployment) bool {
	return d.Name == "vm-import-deployment"
}
```

pkg/operator/controller/controller_suite_test.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+func TestOperator(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Operator Resource Suite")
+}
+
+var testenv *envtest.Environment
+var cfg *rest.Config
+var clientset *kubernetes.Clientset
```
jakub-dzon: `clientset` despite being initilized in `BeforeSuite` is not used.

pkg/operator/controller/controller_suite_test.go
```diff
@@ -0,0 +1,83 @@
...
+var _ = BeforeSuite(func(done Done) {
+	logf.SetLogger(logf.ZapLoggerTo(GinkgoWriter, true))
+
+	env := &envtest.Environment{}
+
+	var err error
+	cfg, err = env.Start()
+	Expect(err).NotTo(HaveOccurred())
+
+	clientset, err = kubernetes.NewForConfig(cfg)
```
jakub-dzon: Is this a kind of sanity check or this is actually not needed?

pkg/operator/resources/operator/operator.go
```diff
@@ -15,12 +15,17 @@ import (
 
 const (
 	operatorName       = "vm-import-operator"
+	controllerName     = "vm-import-controller"
 	serviceAccountName = operatorName
 	roleName           = operatorName
```
jakub-dzon: `roleName` is not used.

tools/csv-generator/csv-generator.go
```diff
@@ -43,12 +45,9 @@ func main() {
 	}
 
 	if *dumpCRDs {
-		vmiCrds := vmioperator.NewCrds()
```
jakub-dzon, line 46: Please remove `NewCrds` from the `operator` package; it does not seem to be used anywhere anymore.

Makefile
machacekondra: It's executed in `cluster/operator-push.sh:33`, but if local image with specific version will be build before locally, it will use it, but it could be confusing.

pkg/operator/resources/operator/operator.go
```diff
@@ -976,23 +1050,126 @@ the disk alias on ovirt DiskMappings is respected only when provided in context
...
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "metadata.name",
+				},
+			},
+		},
+	}
+}
+
+// CreateOperatorDeploymentSpec creates deployment
+func CreateOperatorDeploymentSpec(name, matchKey, matchValue, serviceAccount string, numReplicas int32) *appsv1.DeploymentSpec {
```
jakub-dzon: `name` is unused

jakub-dzon (changes requested):

pkg/operator/controller/controller_suite_test.go
pkliczewski: Thanks

pkg/operator/resources/operator/operator.go
pkliczewski: Thanks

pkg/operator/controller/controller_suite_test.go
pkliczewski: I will remove

tools/csv-generator/csv-generator.go
pkliczewski: It is part of not complete refactoring. Will remove.

pkg/operator/resources/operator/operator.go
pkliczewski: Thanks

pkg/operator/controller/controller.go
pkliczewski: It is not trivial to extract this as method but I will try to reduce the size.

pkg/operator/controller/controller.go
pkliczewski: Same here, number of arguments accepting and returning from this and the above method would be very confusing.

pkg/operator/controller/controller_suite_test.go
pkliczewski: No need to have it, will remove.

pkg/operator/resources/operator/operator.go
pkliczewski: Done

jakub-dzon (approved):

pkg/operator/controller/controller.go
pkliczewski: I checked and it is needed for `envconfig.Process`. Without it we are unable to parse env variable correctly.

pkg/operator/resources/operator/operator.go
pkliczewski: I checked why we have it here and it seems like the logic to `client.Get` and `client.Create` rely on it. We need `ClusterRole` and `ClusterRoleBinding` but we need to handle it by generic code which works with `Deployment`.

masayag: I'm still not sure about it - on line 61 the kind is 'ClusterRole' but the returned type of this method is a Role instead of ClusterRole.

masayag (approved):

PR #204, opened 2020-05-12
Update vm import progress reflect dv import

### LINKED ISSUE #124: Too little percentage of disks import in progress
The total percentage of Disks importing is 50%. Isn't this too little as the import will probably take most of the time?

```go
	progressStart                     = "0"
	progressCreatingVM                = "30"
	progressCopyingDisks              = "40"
	progressStartVM                   = "90"
	progressDone                      = "100"
```

We could also show two variants of VM Import in the UI based on these percentages - so the user would know when the disk import starts.

### LINKED ISSUE #126: disk import progress does not show any change until CDI import completes
depends on: 
* https://github.com/kubevirt/containerized-data-importer/issues/1190
* https://bugzilla.redhat.com/1826381

machacekondra:
Fixes: https://github.com/kubevirt/vm-import-operator/issues/126
Fixes: https://github.com/kubevirt/vm-import-operator/issues/124

Since CDI v1.17.0 imageio progress is updated for the DV import.
This patch changes the update progress of the VM import to following
format:

1) start at 10 if validation is successfull
2) count progress as 10 + (allDvProgress / countOfDisks) * (75 / 100)
3) if all disks import is done we are at 85
4) Start vm 90
5) Finish 100

So for example for two disks of one done on 25% and second for 60%
we go as -> 10 + (85 / 2) * 0.75 = 41.875%

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/kubevirt/vm-import-operator/pkg/config"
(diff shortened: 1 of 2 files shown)


pkliczewski (approved):
Looks good, would be good to add a test or two :)

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -41,12 +42,13 @@ const (
...
 	progressStart                     = "0"
-	progressCreatingVM                = "30"
-	progressCopyingDisks              = "40"
+	progressCreatingVM                = "5"
+	progressCopyingDisks              = "10"
 	progressStartVM                   = "90"
 	progressDone                      = "100"
-	progressForCopyDisk               = 40
 	requeueAfterValidationFailureTime = 5 * time.Second
+	progressForCopyDisk               = 75
```
jakub-dzon: Could you, please, group this with other `progress*` constants?

jakub-dzon (changes requested):

masayag (approved):

jakub-dzon (approved):

PR #207, opened 2020-05-12
Functional test for VM with multus network

jakub-dzon:
```release-note
NONE
```


cluster/common.sh
```diff
@@ -2,6 +2,14 @@
 
 set -e
 
+function install_cnao {
+  ./cluster/kubectl.sh apply -f https://raw.githubusercontent.com/kubevirt/cluster-network-addons-operator/master/manifests/cluster-network-addons/0.35.1/namespace.yaml
```
pkliczewski: It would be great to use a parameter for cluster-network-addons version so we can easily update it in the future.

pkliczewski (approved):

machacekondra (approved):

tests/networking_test.go
```diff
@@ -0,0 +1,81 @@
...
+			Expect(created).To(BeSuccessful(f))
+
+			retrieved, _ := f.VMImportClient.V2vV1alpha1().VirtualMachineImports(namespace).Get(created.Name, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			vmBlueprint := v1.VirtualMachine{ObjectMeta: metav1.ObjectMeta{Name: retrieved.Status.TargetVMName, Namespace: namespace}}
+			Expect(vmBlueprint).To(BeRunning(f))
+
+			By("Having correct network configuration")
+			vm, _ := f.KubeVirtClient.VirtualMachine(namespace).Get(vmBlueprint.Name, &metav1.GetOptions{})
```
masayag, line 67: since the VM is running, why not to examine the VM-Instance ? the status object of it should be updated with data relevant to new interface IIRC

PR #213, opened 2020-05-13
Add variable to change kubevirtci version

machacekondra:
This PR add env variable to change kubevirt ci version if needed. For
now we use master, but we can change it in case there is bug to some
older version.

### DIFF
--- a/cluster/kubevirtci.sh
+++ b/cluster/kubevirtci.sh
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.16'}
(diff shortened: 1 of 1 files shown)


pkliczewski (approved):

cluster/kubevirtci.sh
```diff
@@ -14,13 +14,15 @@
 
 export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.16'}
 
+KUBEVIRTCI_VERSION='master'
```
masayag: wouldn't you like to set 'master' as default or use the value provided by the initiator of the script similar to KUBEVIRT_PROVIDER ?

machacekondra: Not initially, but that's better, thanks.

masayag (approved):

PR #222, opened 2020-05-18
Use new name format of manifest

machacekondra:
We now use new format of name without v2v_v1alpha1 prefix, we should use
it in our tests as well.


pkliczewski (approved):
Thank you!

masayag (approved):
with next release, this will have to be changed as well.

PR #234, opened 2020-05-20
Use new CR for tests

machacekondra:
Change-Id: I3b6ff4af87ea6b5c38b81dbbee0fb95571721230

### DIFF
--- a/cluster/operator-install.sh
+++ b/cluster/operator-install.sh
@@ -17,6 +17,21 @@
 set -ex
 
 export VERSION=v0.0.0
+export VM_IMPORT_CONFIG_STATUS=${VM_IMPORT_CONFIG_STATUS:-''}
 
+# Create and wait for the operator
 ./cluster/kubectl.sh create -f _out/vm-import-operator/${VERSION}/operator.yaml
 ./cluster/kubectl.sh wait deploy/vm-import-operator -n kubevirt --for=condition=Available --timeout=600s
+
+# Create and wait for the controller
+./cluster/kubectl.sh create -f _out/vm-import-operator/${VERSION}/vmimportconfig_cr.yaml
+# When `kubectl wait` will support `--ignore-not` found parameter this `if` can be removed.
+if [[ "$(./cluster/kubectl.sh get deploy/vm-import-controller -n kubevirt 2>&1)" =~ "not found" ]]; then
+    sleep 2
+fi
+./cluster/kubectl.sh wait deploy/vm-import-controller -n kubevirt --for=condition=Available --timeout=600s
+
+if [ ! -z "$VM_IMPORT_CONFIG_STATUS" ]
+then
+  ./cluster/kubectl.sh wait vmimportconfig/vm-import-operator-config --timeout=600s --for=condition=$VM_IMPORT_CONFIG_STATUS
+fi
--- a/cluster/clean.sh
+++ b/cluster/clean.sh
@@ -18,6 +18,5 @@ set -ex
 
 echo 'Cleaning up ...'
 
-./cluster/kubectl.sh delete --ignore-not-found -f _out/vm-import-operator/${VERSION}/resourcemapping_crd.yaml
-./cluster/kubectl.sh delete --ignore-not-found -f _out/vm-import-operator/${VERSION}/virtualmachineimport_crd.yaml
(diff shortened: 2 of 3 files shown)


pkliczewski (approved):

cluster/operator-install.sh
```diff
@@ -19,4 +19,5 @@ set -ex
 export VERSION=v0.0.0
 
 ./cluster/kubectl.sh create -f _out/vm-import-operator/${VERSION}/operator.yaml
+./cluster/kubectl.sh create -f _out/vm-import-operator/${VERSION}/vmimportconfig_cr.yaml
 ./cluster/kubectl.sh wait deploy/vm-import-operator -n kubevirt --for=condition=Available --timeout=600s
```
masayag, line 24: i think the correct condition will be waiting for `deploy/vm-import-controller` to be ready before proceeding with tests - that will indicate the `vmimportconfig_cr` is installed properly and the controller resources where created as expected and ready to serve.

There should be an option to wait on vmimportconfig_cr status a as well

jakub-dzon (approved):

cluster/operator-install.sh
```diff
@@ -17,6 +17,13 @@
 set -ex
 
 export VERSION=v0.0.0
+export VM_IMPORT_CONFIG_STATUS=${VM_IMPORT_CONFIG_STATUS:-''}
 
 ./cluster/kubectl.sh create -f _out/vm-import-operator/${VERSION}/operator.yaml
-./cluster/kubectl.sh wait deploy/vm-import-operator -n kubevirt --for=condition=Available --timeout=600s
+./cluster/kubectl.sh create -f _out/vm-import-operator/${VERSION}/vmimportconfig_cr.yaml
```
masayag, line 27: Ondra, could there be a race and teh vm-import-operator deployment will not be ready and line 24 will fail due to that reason since the vm-import-controller will not be created at that time.
We need to add another wait after line 23 - as you did originally
`./cluster/kubectl.sh wait deploy/vm-import-operator -n kubevirt --for=condition=Available --timeout=600s	`
once it is ready - create the vmimportconfig_cr.yaml
and then wait on vm-import-controller

jakub-dzon: There is a problem:
```
deployment.apps/vm-import-operator created
+ ./cluster/kubectl.sh create -f _out/vm-import-operator/v0.0.0/vmimportconfig_cr.yaml
vmimportconfig.v2v.kubevirt.io/vm-import-operator-config created
+ ./cluster/kubectl.sh wait deploy/vm-import-controller -n kubevirt --for=condition=Available --timeout=600s
Error from server (NotFound): deployments.apps "vm-import-controller" not found
```

masayag: that's why i suggested to add a wait for the vm-import-operator deployment first
then create the vmimport-config cr
and then wait on the vm-import-controller deployment
but there can be a race between the last two - by watching the vmimportconfig status - it should be fine to wait until it is reported to complete it task by making sure the vm-import-controller deployment is ready

machacekondra: > Ondra, could there be a race and teh vm-import-operator deployment will not be ready and line 24 will fail due to that reason since the vm-import-controller will not be created at that time.
> We need to add another wait after line 23 - as you did originally
> `./cluster/kubectl.sh wait deploy/vm-import-operator -n kubevirt --for=condition=Available --timeout=600s `
> once it is ready - create the vmimportconfig_cr.yaml
> and then wait on vm-import-controller

Thanks, fixed.



> There is a problem:
> 
> ```
> deployment.apps/vm-import-operator created
> + ./cluster/kubectl.sh create -f _out/vm-import-operator/v0.0.0/vmimportconfig_cr.yaml
> vmimportconfig.v2v.kubevirt.io/vm-import-operator-config created
> + ./cluster/kubectl.sh wait deploy/vm-import-controller -n kubevirt --for=condition=Available --timeout=600s
> Error from server (NotFound): deployments.apps "vm-import-controller" not found
> ```

Thanks, actually it's bug in kubectl, there is no mechanism to wait on non-existent, there is PR to add `--ignore-not-found`, but not yet released. I'll add some WA.

jakub-dzon (approved):

masayag (approved):

PR #246, opened 2020-05-21
Guidelines

masayag:


CONTRIBUTING.md
```diff
@@ -0,0 +1,116 @@
+# Introduction
+
+VM Import operator requires the following components to be installed:
+
+* **Kubernetes** is a container orchestration system, and is used to run
```
pkliczewski, line 5: When using k8s we would not able able to install templates. I suggest to change it to OKD.

CONTRIBUTING.md
```diff
@@ -0,0 +1,116 @@
...
+  containers on a cluster
+* **containerized-data-importer (CDI)** is an add-on which solves the problem of
+  populating Kubernetes Persistent Volumes with data.  It was written to be
+  general purpose but with the virtualization use case in mind.  Therefore, it
+  has a close relationship and special integration with KubeVirt. Minimal CDI version
+  is v1.15.0 in which oVirt imageio support was added.
+* **KubeVirt** is an add-on which is installed on-top of Kubernetes, to be able
+  to add basic virtualization functionality to Kubernetes.
+* **common-templates** is a set of templates used to create KubeVirt VMs.
```
pkliczewski, line 15: What about network addons + multus?

CONTRIBUTING.md
masayag: I added an explanation below - see line 16:
```
VM Import operator relies on templates to create best-match VM specification.
However, templates aren't k8s nor kubevirt resources, but are part of OpenShift API,
therefore as a base provider for development should be OKD / OpenShift to allow
enhancements of templates.
```

CONTRIBUTING.md
masayag: added.

CONTRIBUTING.md
masayag: +added a note at the same bullet

pkliczewski (approved):

PR #247, opened 2020-05-21
Functional tests for VM validation

jakub-dzon:
```release-note
NONE
```


pkliczewski (approved):

machacekondra:
The random test failure seems to be an issue with this method: https://github.com/kubevirt/vm-import-operator/blob/master/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go#L539 I am investigating.

jakub-dzon:
@machacekondra Is it possible that the problem is with the performance of the environment and timeouts while waiting for expected conditions are too short?

machacekondra (approved):

PR #253, opened 2020-05-26
Update network mapping value for ovirt

masayag:

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -56,7 +56,7 @@ Spec:
         - source:
             name: red/profile1
             target: xyz
-            type: bridge
+            type: multus
         diskMappings: # a mapping of a specific disk to storage class
         - source:
             id: 8181ecc1-5db8-4193-9c92-3ddab3be7b12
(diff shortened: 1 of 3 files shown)


docs/design.md
```diff
@@ -112,7 +112,7 @@ Spec:
       target: xyz
       type: bridge
     - source:
-        name: ovirtmgmt
+        name: ovirtmgmt/ovirtmgmt
```
jakub-dzon: What do you think about making the names reflect that they correspond to vnic profile and network names (i.e -net, -vnic-profile suffixes or something similar)?

masayag: specifically for `ovirtmgmt/ovirtmgmt` I'd like to have it as is, since they are default values in ovirt and changing it to ovirtmgmt-net might be misleading. I'll add a description for that value.
The mapping above it relies to what you suggested `red/profile1`

pkliczewski (approved):

jakub-dzon (approved):

Issue #138, opened 2020-04-22
Propagate label for tracking created resources


masayag:
When vm-import is part of a wider migration, the component that orchestrates multiple migrations can use a label to monitor not only the vmimports CR of the migration, but also the created resources (mainly vms) to gain another level of tracking the progress.

Adding label `vmimport.v2v.kubevirt.io/tracker` on vmimport CR creation will instruct the operator to prorated that label with its value to the created resources (vms)

PR #258, opened 2020-05-27
Add vm description

pkliczewski:
We need to add vm desctiption as defined in BZ.

Bug-Url: https://bugzilla.redhat.com/1838683

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -2,6 +2,7 @@ package virtualmachineimport
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strconv"
@@ -41,6 +42,8 @@ const (
 	sourceVMInitialState = "vmimport.v2v.kubevirt.io/source-vm-initial-state"
 	// AnnCurrentProgress is annotations storing current progress of the vm import
 	AnnCurrentProgress = "vmimport.v2v.kubevirt.io/progress"
+	// AnnPropagate is annotation defining which values to propagate
+	AnnPropagate = "vmimport.v2v.kubevirt.io/propagate-annotations"
 	// constants
 	progressStart         = "0"
 	progressCreatingVM    = "5"
@@ -71,7 +74,7 @@ const (
 )
 
 var (
-	log                          = logf.Log.WithName("controller_virtualmachineimport")
+	log = logf.Log.WithName("controller_virtualmachineimport")
 	// importPodRestartTolerance define how many restart of the import pod are tolerated before
 	// we end the import as failed, by default it's 3.
 	importPodRestartTolerance, _ = strconv.Atoi(os.Getenv("IMPORT_POD_RESTART_TOLERANCE"))
@@ -425,6 +428,9 @@ func (r *ReconcileVirtualMachineImport) createVM(provider provider.Provider, ins
 		return "", err
 	}
 
+	// propagate annotations
+	setAnnotations(instance, vmSpec)
+
 	// Update progress:
 	if err = r.updateProgress(instance, progressStart); err != nil {
 		return "", err
@@ -484,6 +490,26 @@ func (r *ReconcileVirtualMachineImport) createVM(provider provider.Provider, ins
 	}
 
 	return found.Name, nil
+}
+
+func setAnnotations(instance *v2vv1alpha1.VirtualMachineImport, vmSpec *kubevirtv1.VirtualMachine) {
+	reqLogger := log.WithValues("Request.Namespace", instance.Namespace, "Request.Name", instance.Name)
+	annotations := instance.GetAnnotations()
+	propagate := annotations[AnnPropagate]
+	if propagate != "" {
+		annotations := vmSpec.GetAnnotations()
+		if annotations == nil {
+			annotations = map[string]string{}
+			vmSpec.SetAnnotations(annotations)
+		}
+		prop := map[string]string{}
+		err := json.Unmarshal([]byte(propagate), &prop)
+		if err != nil {
+			reqLogger.Info("Failed while parsing annotations to propagate to virtual machine")
+			return
+		}
+		utils.WithLabels(annotations, prop)
+	}
 }
 
 // startVM start the VM if was requested to be started and VM disks are imported and ready:
--- a/examples/ovirt_virtualmachineimport_example_cr.yaml
+++ b/examples/ovirt_virtualmachineimport_example_cr.yaml
@@ -5,6 +5,7 @@ metadata:
   namespace: example-ns
   annotations:
     vmimport.v2v.kubevirt.io/progress: "20"
+    vmimport.v2v.kubevirt.io/propagate-annotations: "{'description': 'My description.'}"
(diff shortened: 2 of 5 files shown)


pkliczewski:
This PR provides vm description capabilities together with https://github.com/ManageIQ/manageiq-v2v-conversion_host/pull/64

pkliczewski:
@suomiy FYI

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -425,6 +427,16 @@ func (r *ReconcileVirtualMachineImport) createVM(provider provider.Provider, ins
 		return "", err
 	}
 
+	if instance.Spec.VMDescription != nil {
+		desc := utils.MakeLabelFrom(*instance.Spec.VMDescription)
```
jakub-dzon: Shortening is not needed here unless there is a business requirement - the limit of 63 characters is imposed on the annotation name and on labels.
(reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/).

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -41,6 +41,8 @@ const (
 	sourceVMInitialState = "vmimport.v2v.kubevirt.io/source-vm-initial-state"
 	// AnnCurrentProgress is annotations storing current progress of the vm import
 	AnnCurrentProgress = "vmimport.v2v.kubevirt.io/progress"
+	// AnnVMDesctiption is annotation storing vm description
```
masayag: s/AnnVMDesctiption/AnnVMDescription

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
```diff
@@ -22,6 +22,9 @@ type VirtualMachineImportSpec struct {
 	// +optional
 	TargetVMName *string `json:"targetVmName,omitempty"`
 
+	// +optional
+	VMDescription *string `json:"vmDescription,omitempty"`
```
atiratree: should we consider renaming the fields so they do not include `vm` inside? 

Especially in case if we end up importing templates (not sure about vmware part) as well?

masayag: I can suggest an alternate method for it: instead of including it as part of the API, we can define annotation that will contain a list of key-values map as its value and its purpose will be to propagate it to the target vm, without the need to specify the name of the fields explicitly. 

in this case, we can agree that if there is an annotation named `vmimport.v2v.kubevirt.io/propagate-annotations` - its values will be  propagated to target VM

An example of it in a vm-import-cr could look like:
```
apiVersion: v2v.kubevirt.io/v1alpha1
kind: VirtualMachineImport
metadata:
  name: example-virtualmachineimport
  namespace: example-ns
  annotations:
    vmimport.v2v.kubevirt.io/propagate-annotations:
      "description": "the-resource-description",
      "another-annotation": "...",
...
```
@pkliczewski what do you think ?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: Thanks, will remove.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: :+1:

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
pkliczewski: That was something I wanted to do at the beginning but @suomiy suggested to have it in spec. We could have common logic to pass annotations and labels and in this way we could fix https://github.com/kubevirt/vm-import-operator/issues/138 as well. @suomiy what do you think?

atiratree: I agree something like this would be very useful - also for the labels.

anyway I am not sure the implementation you are proposing is feasible.

The annotations are expected to be mapping of string to string. Not sure if you can change it in your definition, but all other projects/resources expected this to be a string.

Maybe we could create a new field for this?

pkliczewski: We could use json string instead. It should not be that difficult to create it in the UI. @masayag what do you think?

masayag: It can simply be multi-line value to be parsed into a map, my example should be changed to contain the multi-line notation by adding | after `    vmimport.v2v.kubevirt.io/propagate-annotations:
`

atiratree: Yes, not difficult for the UI. Just harder for people trying to use YAML

pkliczewski: I do not think it would be used from command line very heavily.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -41,6 +42,10 @@ const (
 	sourceVMInitialState = "vmimport.v2v.kubevirt.io/source-vm-initial-state"
 	// AnnCurrentProgress is annotations storing current progress of the vm import
 	AnnCurrentProgress = "vmimport.v2v.kubevirt.io/progress"
+	// AnnVMDescription is annotation storing vm description
+	AnnVMDescription = "description"
```
masayag: this can be removed.

pkliczewski: Done

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -486,6 +494,24 @@ func (r *ReconcileVirtualMachineImport) createVM(provider provider.Provider, ins
...
+	if propagate != "" {
+		annotations := vmSpec.GetAnnotations()
+		if annotations == nil {
+			annotations = map[string]string{}
+			vmSpec.SetAnnotations(annotations)
+		}
+		prop := map[string]string{}
+		err := json.Unmarshal([]byte(propagate), &prop)
+		if err != nil {
+			return
```
masayag, line 509: pls log the error, i.e. failed to parse annotations in case of an error before returning.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
masayag: Also, the user will be able to edit the VM resource after it was created instead of relying on the vm-import process to do so.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: It is not critical so Info level should be good enough.

masayag (approved):

jakub-dzon (approved):

machacekondra (approved):

PR #261, opened 2020-05-28
Reduce restrictions on targer in mappings

pkliczewski:
We allow for target not to be there and we would use defaults.


pkliczewski:
@suomiy FYI

jakub-dzon (approved):

machacekondra (approved):

masayag:
@jakub-dzon could you update design doc about this change ? for each type disk-mapping/storage-mapping/network-mapping what is assumed to be the defaults to be set by the controller.

jakub-dzon:
Sure, I will.

PR #262, opened 2020-05-28
Bump default k8s version to 1.18

machacekondra:
And also use integrated CNAO, instead of explicit installation.

### DIFF
--- a/cluster/common.sh
+++ b/cluster/common.sh
@@ -2,15 +2,6 @@
 
 set -e
 
-function install_cnao {
-  CNAO_VERSION=${CNAO_VERSION:-'0.35.1'}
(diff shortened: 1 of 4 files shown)


cluster/common.sh
```diff
@@ -2,15 +2,6 @@
 
 set -e
 
-function install_cnao {
```
pkliczewski, line 5: Why we do not need this anymore?

machacekondra: It's part of the `kubevirtci` repo [now](https://github.com/kubevirt/kubevirtci/tree/master/cluster-up/cluster/k8s-1.17#bringing-the-cluster-up-with-cluster-network-addons-operator-provisioned). It's enabled just by passing `KUBEVIRT_WITH_CNAO=true`

machacekondra: To make it effect, this must be merged before: https://github.com/kubevirt/project-infra/pull/449

pkliczewski: OK, thanks

pkliczewski (approved):

jakub-dzon (approved):

PR #274, opened 2020-06-02
Set unique data-volume name

### LINKED ISSUE #165: vm is missing disks
When importing VM second time (meaning DV and PVC already exists) the import completes immediately, but the VM is created without any disks

1. The VMs shouldn't point to the same DV/PVC as one might do some changes on one and then decide to import the VM in previous state from oVirt
2. would be good to add VM name to the DV and append random string (generateName)  to the DV/PVC

masayag:
Setting data-volume name as `target-vm-name + "-" + disk-attachment-id`
will assure that the DV name is unique.
This will allow importing the same disk or the same vm with a different
target name.

Using 'GenerateName' option is problematic, since there is certain
logic that relies on a known and consistent DV Name prior to its
creation.

Fixes #165

### DIFF
--- a/pkg/providers/ovirt/mapper/mapper.go
+++ b/pkg/providers/ovirt/mapper/mapper.go
@@ -199,7 +199,7 @@ func (o *OvirtMapper) MapDisks(vmSpec *kubevirtv1.VirtualMachine, dvs map[string
 	disks := make([]kubevirtv1.Disk, len(dvs))
 	for id := range dvs {
 		diskAttachments, _ := o.vm.DiskAttachments()
-		diskAttachment := getDiskAttachmentByID(id, diskAttachments)
+		diskAttachment := getDiskAttachmentByID(id, diskAttachments, vmSpec.ObjectMeta.Name)
 		iface, _ := diskAttachment.Interface()
 		disks[i] = kubevirtv1.Disk{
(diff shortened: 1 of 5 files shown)


pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -247,13 +248,13 @@ func (o *OvirtMapper) MapDataVolumes() (map[string]cdiv1.DataVolume, error) {
...
 
-		dvs[attachID] = cdiv1.DataVolume{
+		dvs[dvName] = cdiv1.DataVolume{
 			TypeMeta: metav1.TypeMeta{
 				APIVersion: cdiAPIVersion,
 				Kind:       dataVolumeKind,
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      attachID,
+				Name:      dvName,
```
pkliczewski, line 257: Why not to use `GenerateName` here? We could use myattchID- and k8s would make sure it is unique. In `dvs` slice we could use myattchID as a key and find dv using `strings.HasPrefix`

pkliczewski: Thinking about it more we can use target vm name as a prefix.

pkliczewski (approved):

machacekondra (approved):

PR #287, opened 2020-06-11
Refactor importDisks method

machacekondra:
This PR refactor importDisks, so we don't cycle over all DVs twice.

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -296,7 +296,7 @@ func (r *ReconcileVirtualMachineImport) importDisks(provider provider.Provider, 
 
 	dvsDone := make(map[string]bool)
 	dvsImportProgress := make(map[string]float64)
-	for dvID := range dvs {
+	for dvID, dv := range dvs {
 		if err = r.addWatchForImportPod(instance, dvID); err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -635,54 +641,57 @@ func (r *ReconcileVirtualMachineImport) createDataVolumes(provider provider.Prov
...
+	r.recorder.Eventf(
+		instance,
+		corev1.EventTypeNormal,
+		EventImportInProgress,
+		"Import of Virtual Machine %s/%s disk %s in progress", vmName.Namespace, vmName.Name, dv.Name,
+	)
 
 	// Update datavolume in VM import CR status:
-	if err = r.updateDVs(instanceNamespacedName, dvs); err != nil {
+	if err = r.updateDVs(instanceNamespacedName, dv); err != nil {
```
masayag, line 689: @machacekondra do you see any option for this method or `updateVMSpecDataVolumes` be called more than once for the created DV ? Any probability of `updateDVs` to succeed and `updateVMSpecDataVolumes` to fail ?

I'd like to make sure this method will not be called more than once per each DV and to make sure the list of reported DV for the vmimport CR's status is updated correctly.

machacekondra: Well, I can't think of any case, when it can happen, but I will improve those two calls, so it's more reliable, I've opened an [issue](https://github.com/kubevirt/vm-import-operator/issues/290), and will fix in separate PR.

masayag (approved):

pkliczewski (approved):

jakub-dzon (approved):

PR #289, opened 2020-06-15
New negative test cases descriptions

jakub-dzon:
```release-notes
NONE
```

### DIFF
--- a/docs/functional-tests.md
+++ b/docs/functional-tests.md
@@ -13,8 +13,16 @@
 ## Basic VM negative
 | Test description | Implemented |
 | :---------------- | :-----------: |
-| Basic VM import with invalid VM image | &check; |
-| Basic VM import with invalid VM image size | &check; |
+| Basic VM import with invalid VM image should fail import | &check; |
+| Basic VM import with invalid VM image size should fail import | &check; |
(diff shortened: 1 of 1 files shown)


pkliczewski (approved):

docs/functional-tests.md
```diff
@@ -13,8 +13,16 @@
 ## Basic VM negative
 | Test description | Implemented |
 | :---------------- | :-----------: |
-| Basic VM import with invalid VM image | &check; |
-| Basic VM import with invalid VM image size | &check; |
+| Basic VM import with invalid VM image should fail import | &check; |
+| Basic VM import with invalid VM image size should fail import | &check; |
```
masayag, line 17: what 'invalid' stands for ? wrong value instead of number or size too big for storage class to provide ?

jakub-dzon: From the looks of the test that was written it's a number that does not correspond to the actual image size (i think). 
@machacekondra - do you confirm that?

machacekondra: Right, that's actually the test case that we properly handle when CDI pod keeps crashing, and it end with `crashbackoflopp` state

masayag (approved):

PR #296, opened 2020-06-18
Propagate env variables to vm-import-controller

masayag:


pkg/operator/controller/controller.go
```diff
@@ -752,6 +757,22 @@ func (r *ReconcileVMImportConfig) getOperatorArgs(cr *vmimportv1alpha1.VMImportC
 		}
 	}
 
+	// update operatorArgs with OS config map name and namespace if defined in vm-import-operator deployment
+	// TODO: Replace vm-import-operator deployment env vars with VMImportConfig attributes
```
pkliczewski, line 761: Let's update VMImportConfig only in master. The bug is targeted to 2.5

pkg/operator/controller/controller.go
```diff
@@ -752,6 +757,22 @@ func (r *ReconcileVMImportConfig) getOperatorArgs(cr *vmimportv1alpha1.VMImportC
 		}
 	}
 
+	// update operatorArgs with OS config map name and namespace if defined in vm-import-operator deployment
```
pkliczewski, line 760: Do you want to customize it in csv generator?

jakub-dzon (approved):

pkg/operator/controller/controller.go
masayag: I'll just add those env-vars definition, but providing a value for them should be done by a user if needed.

pkliczewski (approved):

PR #298, opened 2020-06-19
Functional test for source VM that wouldn't go down

jakub-dzon:
```release-notes
NONE
```

### DIFF
--- a/tests/ovirt/basic_vm_import_negative_test.go
+++ b/tests/ovirt/basic_vm_import_negative_test.go
@@ -1,24 +1,28 @@
 package ovirt_test
 
 import (
+	"time"
+
+	"github.com/kubevirt/vm-import-operator/tests/ovirt"
+	"github.com/onsi/ginkgo/extensions/table"
+	"k8s.io/apimachinery/pkg/api/errors"
+	v1 "kubevirt.io/client-go/api/v1"
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
+
 	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
+	"github.com/kubevirt/vm-import-operator/pkg/conditions"
 	fwk "github.com/kubevirt/vm-import-operator/tests/framework"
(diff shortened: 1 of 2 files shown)


pkliczewski (approved):

tests/ovirt/basic_vm_import_negative_test.go
```diff
@@ -134,9 +138,44 @@ var _ = Describe("VM import", func() {
...
+
+		Consistently(func() (*v2vv1alpha1.VirtualMachineImportCondition, error) {
+			retrieved, err := f.VMImportClient.V2vV1alpha1().VirtualMachineImports(f.Namespace.Name).Get(created.Name, metav1.GetOptions{})
+			if err != nil {
+				return nil, err
+			}
+			condition := conditions.FindConditionOfType(retrieved.Status.Conditions, v2vv1alpha1.Processing)
+			return condition, nil
+
+		}, 5*time.Minute, time.Minute).Should(BeNil())
```
machacekondra, line 174: Should we also check that after the timeout, we get the error?

machacekondra (approved):

tests/ovirt/basic_vm_import_negative_test.go
jakub-dzon: What error do you have in mind? The timeout here only specifies how long the assertion should be kept, it does not return error after the timeout. In this case it's 5 minutes, which is as long as the ovirt client waits for the VM to be down in one reconciliation and that long I wanted to make sure that it doesn't start processing.

machacekondra: I mean this one: https://github.com/kubevirt/vm-import-operator/blob/master/pkg/providers/ovirt/client/ovirt-client.go#L139

When the timeout (5min) is out, the stopVm method should return error.

jakub-dzon: It does but on high level it is translated only to reconciliation error and CR re-queuing (https://github.com/kubevirt/vm-import-operator/blob/12de4cca49c113ce028fcb84934a64541009922e/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go#L239); nothing we can detect except for verifying it doesn't progress beyond mapping rules verification.

PR #299, opened 2020-06-19
VM import CR removal after successful import - e2e test

jakub-dzon:
```release-note
NONE
```

### DIFF
--- a/tests/ovirt/basic_vm_import_test.go
+++ b/tests/ovirt/basic_vm_import_test.go
@@ -1,15 +1,17 @@
 package ovirt_test
 
 import (
+	"time"
+
 	v2vv1alpha1 "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1alpha1"
 	"github.com/kubevirt/vm-import-operator/tests"
-	"github.com/kubevirt/vm-import-operator/tests/ovirt/vms"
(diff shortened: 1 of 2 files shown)


pkliczewski (approved):

tests/ovirt/basic_vm_import_test.go
```diff
@@ -109,6 +111,61 @@ var _ = Describe("Basic VM import ", func() {
...
+				deleteOptions := metav1.DeleteOptions{
+					PropagationPolicy: &foreground,
+				}
+				err = vmImports.Delete(retrieved.Name, &deleteOptions)
+				if err != nil {
+					Fail(err.Error())
+				}
+
+				By("Waiting for VM import removal")
+				err = f.EnsureVMIDoesNotExist(retrieved.Name)
```
machacekondra, line 144: Just a light note, that I am sometimes confused wheter `VMI` is `VirtualMachineInstance` or `VirtualMachineImport`, maybe we should rename those funtions to be clear what it handle.

machacekondra (approved):

tests/ovirt/basic_vm_import_test.go
jakub-dzon: Good point; I will change it in the next PR.

PR #302, opened 2020-06-24
CNV-5368: Failing import when template is not found

jakub-dzon:
Check to be performed when dedicated feature gate (`ImportWithoutTemplate`) is enabled.

```release-note
NONE
```

### DIFF
--- a/cluster/common.sh
+++ b/cluster/common.sh
@@ -18,7 +18,7 @@ function install_cdi {
 function install_kubevirt {
     export KUBEVIRT_VER=$(curl -s https://github.com/kubevirt/kubevirt/releases/latest | grep -o "v[0-9]\.[0-9]*\.[0-9]*")
     ./cluster/kubectl.sh apply -f https://github.com/kubevirt/kubevirt/releases/download/${KUBEVIRT_VER}/kubevirt-operator.yaml
-    ./cluster/kubectl.sh create configmap -n kubevirt kubevirt-config --from-literal=feature-gates=DataVolumes
+    ./cluster/kubectl.sh create configmap -n kubevirt kubevirt-config --from-literal=feature-gates=DataVolumes,ImportWithoutTemplate
     ./cluster/kubectl.sh apply -f https://github.com/kubevirt/kubevirt/releases/download/${KUBEVIRT_VER}/kubevirt-cr.yaml
     ./cluster/kubectl.sh -n kubevirt wait kv kubevirt --for condition=Available --timeout=1200s || debug kubevirt
 }
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -428,16 +428,36 @@ func (r *ReconcileVirtualMachineImport) createVM(provider provider.Provider, ins
 	targetVMName := mapper.ResolveVMName(instance.Spec.TargetVMName)
 
 	// Define VM spec
+	// Update condition to VM template matching:
+	processingCond := conditions.NewProcessingCondition(string(v2vv1alpha1.VMTemplateMatching), "Matching virtual machine template", corev1.ConditionTrue)
+	if err := r.upsertStatusConditions(instanceNamespacedName, processingCond); err != nil {
+		return "", err
+	}
 	template, err := provider.FindTemplate()
 	var spec *kubevirtv1.VirtualMachine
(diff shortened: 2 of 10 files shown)


pkliczewski (approved):

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -428,16 +428,34 @@ func (r *ReconcileVirtualMachineImport) createVM(provider provider.Provider, ins
...
+	}
 	template, err := provider.FindTemplate()
 	var spec *kubevirtv1.VirtualMachine
+	config, cfgErr := r.kvConfigProvider.GetConfig()
+	if cfgErr != nil {
+		log.Error(cfgErr, "Cannot get KubeVirt cluster config.")
+	}
 	if err != nil {
 		reqLogger.Info("No matching template was found for the virtual machine. Using empty VM definition")
+		if !config.ImportWithoutTemplateEnabled() {
```
masayag, line 444: The message above should be split into 2 parts, the first part - should remain there as:
`reqLogger.Info("No matching template was found for the virtual machine.")`

and the second part after this condition ends successfully, before line 449:
`reqLogger.Info("Using empty VM definition")`

The reason is to avoid the message of "Using empty VM definition" if this isn't supported by the setup.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -428,16 +428,34 @@ func (r *ReconcileVirtualMachineImport) createVM(provider provider.Provider, ins
...
+			}
+			return "", err
+		}
 		spec = mapper.CreateEmptyVM(targetVMName)
 	} else {
 		reqLogger.Info("A template was found for creating the virtual machine", "Template.Name", template.ObjectMeta.Name)
 		spec, err = provider.ProcessTemplate(template, targetVMName, instance.Namespace)
 		if err != nil {
 			reqLogger.Info("Failed to process the template. Using empty VM definition. Error: " + err.Error())
+			if !config.ImportWithoutTemplateEnabled() {
```
masayag, line 457: same as above. Let's first just record the failure of processing the template and after the condition log the message of using empty VM definition.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
jakub-dzon: Done.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
jakub-dzon: Done

machacekondra (approved):

cluster/common.sh
```diff
@@ -18,7 +18,7 @@ function install_cdi {
 function install_kubevirt {
     export KUBEVIRT_VER=$(curl -s https://github.com/kubevirt/kubevirt/releases/latest | grep -o "v[0-9]\.[0-9]*\.[0-9]*")
     ./cluster/kubectl.sh apply -f https://github.com/kubevirt/kubevirt/releases/download/${KUBEVIRT_VER}/kubevirt-operator.yaml
-    ./cluster/kubectl.sh create configmap -n kubevirt kubevirt-config --from-literal=feature-gates=DataVolumes
+    ./cluster/kubectl.sh create configmap -n kubevirt kubevirt-config --from-literal=feature-gates=DataVolumes,ImportWithoutTemplate
```
machacekondra, line 21: Not sure if it's set in stone, but maybe we can also consider to decide whether to import based on if the template CRD exists, so we will always fail on openshift, and we will always continue on k8s. And we don't need any config option.

jakub-dzon: With this approach we won't be able to implement functional test for the Openshift behaviour when the template is missing (failing imports). Also, turning this feature off in Openshift would be more difficult and sometimes impossible - someone might want to be able to import VM without a template and use templates for other VMs.

masayag: I like the approach that a feature is disabled by default and should be enabled explicitly for users that decided to use it.
I see a value in documenting this flag and how to enable it in the design document.

masayag (approved):

cluster/common.sh
jakub-dzon: :+1: I've just pushed the change in the documentation.

masayag (approved):

PR #303, opened 2020-06-24
Using pod network type as default when type is missing in the resource mapping

### LINKED ISSUE #301: Validate network target type in network mapping
Having the validation will help users use the API correctly.

Only `pod` and `multus` should be allowed.

jakub-dzon:
Fixes #301.
Fixes:
- https://bugzilla.redhat.com/show_bug.cgi?id=1850467
- https://bugzilla.redhat.com/show_bug.cgi?id=1850425

```release-note
NONE
```


machacekondra (approved):

pkliczewski (approved):

tests/common-vars.go
```diff
@@ -11,4 +11,6 @@ var (
 	PodType = "pod"
 	// MultusType defines `pod` resource mapping network type
 	MultusType = "multus"
+	// UnsupportedType defines non-existing, unsupported `unsupported resource mapping network type
```
masayag, line 14: @jakub-dzon should the second "`unsupported" be removed from the description ? also there is a missing comma.

jakub-dzon: The second "unsupported" is the the value used as unsupported type:) I will fix the missing punctuation mark.

Issue #267, opened 2020-06-01
[Tests] LiveMigratable strategy doesn't work with non-shared pvcs


jakub-dzon:
VMI dies during functional test importing VM that has `LiveMigratable` EvictionStrategy set:
```
 Conditions:
    Last Probe Time:       <nil>
    Last Transition Time:  <nil>
    Message:               cannot migrate VMI with non-shared PVCs
    Reason:                DisksNotLiveMigratable
    Status:                False
    Type:                  LiveMigratable
```

PR #307, opened 2020-06-25
Set PVC accessMode based on source VM

machacekondra:
This PR set the access mode of the DataVolume's PVC based on the source
VM configuration. If the source VM disk is readonly we create read-only
PVC. If the source VM has migratable affinity set, we create
ReadWriteMany PVC.

### DIFF
--- a/pkg/providers/ovirt/mapper/mapper.go
+++ b/pkg/providers/ovirt/mapper/mapper.go
@@ -218,6 +218,22 @@ func (o *OvirtMapper) mapDiskInterface(iface ovirtsdk.DiskInterface) string {
 	return string(iface)
 }
 
+// Set the access mode of the PVC based on the VM's disk read only attribute
+// and based on the affinity settings of the VM.
+func (o *OvirtMapper) getAccessMode(diskAttachment *ovirtsdk.DiskAttachment) corev1.PersistentVolumeAccessMode {
+	accessMode := corev1.ReadWriteOnce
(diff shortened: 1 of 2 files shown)


pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -258,7 +275,7 @@ func (o *OvirtMapper) MapDataVolumes(targetVMName *string) (map[string]cdiv1.Dat
 				},
 				PVC: &corev1.PersistentVolumeClaimSpec{
 					AccessModes: []corev1.PersistentVolumeAccessMode{
-						corev1.ReadWriteOnce,
+						accessMode,
```
masayag, line 278: will this require changes to the functional tests setup to make sure PVs with proper access modes are available or the default PVs supports all types of access-modes ?

machacekondra: Actually, you can create `ReadWriteMany` PVC even on storageclass, which doesn't support RWX acccess mode. So it doesn't but I will add `rook-ceph` in other PR, so we have real test case, and also we may test multiple storage classes.

pkliczewski (approved):

jakub-dzon (approved):

masayag (approved):

machacekondra:
Fixes: https://github.com/kubevirt/vm-import-operator/issues/267

PR #310, opened 2020-06-28
Do not requeue new request if provider init failed

masayag:
In case of new request, do not reuque the request if the provider
initialization failed due to:
* Missing secret or invalid secret format
* Bad credentials or inability to connect to the provider

In this case, the vmimport will be failed, a condition of type Succeeded
will be created with ValidationFailed reason.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1849527

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -213,9 +213,17 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-	err = r.initProvider(instance, provider)
+	message, err := r.initProvider(instance, provider)
 	if err != nil {
-		return reconcile.Result{}, err
+		if r.vmImportInProgress(instance) {
+			return reconcile.Result{}, err
+		}
+		// fail new request and don't requeue it if the provider couldn't be initialized properly
+		err = r.failNewImportProcess(instance, message, err)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
 	}
 	defer provider.Close()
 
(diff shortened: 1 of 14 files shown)


pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -215,7 +215,18 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 	}
 	err = r.initProvider(instance, provider)
 	if err != nil {
-		return reconcile.Result{}, err
+		if r.vmImportInProgress(instance) {
```
pkliczewski, line 218: What do you think about extracting this code to a method?

pkg/providers/ovirt/provider.go
```diff
@@ -155,6 +155,14 @@ func (o *OvirtProvider) Init(secret *corev1.Secret, instance *v2vv1alpha1.Virtua
 	if len(o.ovirtSecretDataMap["caCert"]) == 0 {
 		return fmt.Errorf("oVirt secret caCert cannot be empty")
 	}
+	client, err := o.getClient()
```
pkliczewski, line 182: In the past we wanted to have lazy client to limit number of the connections if not needed. It would be great if we could create `TestConnection` method otherwise we would test the connection in every loop we make.

pkliczewski:
It looks good to me. Let's make prow happy and it is good to merge.

tests/ovirt/basic_vm_import_negative_test.go
```diff
@@ -85,7 +85,7 @@ var _ = Describe("VM import", func() {
 		created, err := f.VMImportClient.V2vV1alpha1().VirtualMachineImports(namespace).Create(&vmi)
 
 		Expect(err).NotTo(HaveOccurred())
-		Expect(created).To(HaveValidationFailure(f, string(v2vv1alpha1.SecretNotFound)))
+		Expect(created).To(BeUnsuccessful(f, string(v2vv1alpha1.ValidationFailed)))
```
masayag, line 88: @jakub-dzon I ended up just with `BeUnsuccessful` instead maintaining the previous assertion `HaveValidationFailure` since updating the status.condition isn't deterministic. In order not to block this PR, I'm using `BeUnsuccessful`, but we should explore why the status.conditions is being overridden in certain cases only by the Successful condition type, and in other cases the Valid condition type remain as well.

pkliczewski (approved):

tests/ovirt/basic_vm_import_negative_test.go
jakub-dzon: Could you, please, create an issue to track that?

jakub-dzon (approved):

tests/ovirt/basic_vm_import_negative_test.go
masayag: Done - https://github.com/kubevirt/vm-import-operator/issues/322

Issue #292, opened 2020-06-15
Add support for importing from VMWare


mansam:
I'm going to work on implementing support for importing VMs from vSphere. As far as I can tell from studying the VM Import Operator code, it's going to involve at least the following things:

1. Update VirtualMachineImport CRD to support a "vmware" key
2. Determine minimum information required to uniquely identify a VM for migration.
3. Possibly refactor where various interfaces live or required methods on interfaces
   * e.g. move SourceClientFactory into client/client.go instead of in the ovirt client package
4. Implement Vmware provider
	* Implement rich client
	* Implement resource mapper
	* Implement template/os finder
5. Make any necessary changes to the virtualmachineimport_controller to support Vmware provider
6. Determine Vmware import validation rules
7. Tests

Before I get too far into the weeds, is there anything it looks like I'm missing or that I should know?

masayag:
Hi @mansam,

- You should also define vmware secret section in the secret resource:
  https://github.com/kubevirt/vm-import-operator/blob/master/docs/design.md#provider-secret
- Importing an image from vmware will require adding support for it on CDI side. Please add a reference to that issue as well.
- Regarding tests, there are two level, unit-tests and functional-tests. For functional tests there is an emulation of ovirt provider at https://github.com/machacekondra/fakeovirt
- Update design documents with the required changes for vmware + add examples of using the new 'vmware' key 
- We'd like to minimize provider specific logic in virtualmachineimport_controller

pkliczewski:
All the PRs are merged. Closing

PR #313, opened 2020-06-29
Add VMware client

mansam:
This PR adds a VMware client to be used as part of a VMware source for VM import, along with tests. It also refactors the SourceClientFactory to be in the client package where it can be shared between providers.

See https://github.com/kubevirt/vm-import-operator/issues/292


pkg/client/client.go
```diff
@@ -12,3 +19,34 @@ type VMClient interface {
 	StartVM(id string) error
 	Close() error
 }
+
```
pkliczewski, line 22: This should stay in the provider code. In this package we should only have interfaces.

pkg/providers/vmware/client/client.go
```diff
@@ -0,0 +1,86 @@
...
+	"context"
+	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/types"
+	"net/url"
+)
+
+// RichOvirtClient is responsible for retrieving VM data from oVirt API
+type richVmwareClient struct {
+	client *govmomi.Client
```
pkliczewski: Why not to add `context` to this struct so it can be used for all interactions with the govmomi client?

pkg/providers/vmware/client/client.go
```diff
@@ -0,0 +1,86 @@
...
+	powerState, err := vm.PowerState(context.TODO())
+	if err != nil {
+		return err
+	}
+	if powerState != types.VirtualMachinePowerStatePoweredOff {
+		task, err := vm.PowerOff(context.TODO())
+		if err != nil {
+			return err
+		}
+		return task.Wait(context.TODO())
```
pkliczewski: For how long this will `wait`? Please not that we do not want to block reconcile loop. There could be other CRs which require processing.

pkg/providers/vmware/client/client.go
```diff
@@ -0,0 +1,86 @@
...
+	powerState, err := vm.PowerState(context.TODO())
+	if err != nil {
+		return err
+	}
+	if powerState != types.VirtualMachinePowerStatePoweredOn {
+		task, err := vm.PowerOn(context.TODO())
+		if err != nil {
+			return err
+		}
+		return task.Wait(context.TODO())
```
pkliczewski: Same here

pkg/providers/vmware/client/client_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"github.com/kubevirt/vm-import-operator/pkg/providers/vmware/client"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/simulator"
+)
+
+var _ = Describe("Test VMware rich client", func() {
```
pkliczewski, line 12: Please think how to mock vmware for functional tests. We created fakeovirt and I think you need something similar.

pkg/client/client.go
mansam: The `newReconciler` function was doing `factory := ovirtprovider.NewSourceClientFactory()` directly. It made sense to me to move `SourceClientFactory` and `NewSourceClientFactory` into a shared place so that `newReconciler` could do `client.NewSourceClientFactory()`, which would support both providers. How would you recommend I do this instead?

pkg/providers/vmware/client/client_test.go
mansam: I believe the simulator package in the vmware client library (`govmomi`) can handle that. I'll continue to investigate it.

pkg/client/client.go
pkliczewski: Thinking a bit more about it we can move it here as you suggested.

pkg/providers/vmware/client/client.go
mansam: @pkliczewski I've updated these operations to use a context with a timeout. Do you have any thoughts on appropriate timeout length?

pkg/client/client.go
```diff
@@ -13,3 +20,34 @@ type VMClient interface {
...
+		URL:      dataMap["apiUrl"],
+		Username: dataMap["username"],
+		Password: dataMap["password"],
+		CACert:   []byte(dataMap["caCert"]),
+	})
+}
+
+// NewVmwareClient creates new VMWare clients
+func (f *SourceClientFactory) NewVmwareClient(dataMap map[string]string) (VMClient, error) {
+	insecure, err := strconv.ParseBool(dataMap["insecure"])
```
machacekondra: Can you add support for the secure method? And drop the insecure?

pkg/providers/vmware/client/client.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+	"net/url"
+	"time"
+)
+
+// timeout value in seconds for vmware api requests
+const timeout = 5 * time.Second
```
machacekondra: Is this timeout OK for any vmware request, or can some request take more time? I just wonder whether this should be somehow configurable for some slow envs.

mansam: I'm not sure; I was hoping to get some feedback from the team on the length and whether it should be configurable.

pkliczewski: Some of the envs could be slower. Are we going to retry if a request would time out?

pkg/client/client.go
mansam: I'm not sure I understand the question. The client is secure by default and only is insecure if user specifies `insecure=true` in order to disable ssl verification. I'll make sure I handle the case where `insecure` isn't present in the dataMap at all to make sure I've got it defaulting to secure.

pkg/providers/vmware/client/client.go
mansam: I was thinking the context error that gets propagated up to Reconcile would result in a requeue (so it wouldn't block the reconcile for other resources, but would still retry via a requeue), but I may be mistaken about that.

pkg/client/client.go
machacekondra: Yes, I've meant add support to CA cert, and don't use insecure at all actually. We don't support `insecure` for oVirt as well.

mansam: Ah, understood. Can do.

pkg/providers/vmware/client/client.go
```diff
@@ -0,0 +1,145 @@
...
+	"time"
+)
+
+// timeout value in seconds for vmware api requests
+const timeout = 30 * time.Second
+
+// RichOvirtClient is responsible for retrieving VM data from oVirt API
+type RichVmwareClient struct {
+	client         *vim25.Client
+	url            *url.URL
```
jakub-dzon: It doesn't look like the whole URL is needed; what do you think about using `url.Userinfo` instead?

pkg/providers/vmware/client/client.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/vmware/govmomi/vim25/soap"
+	"github.com/vmware/govmomi/vim25/types"
+	"net/url"
+	"time"
+)
+
+// timeout value in seconds for vmware api requests
+const timeout = 30 * time.Second
+
+// RichOvirtClient is responsible for retrieving VM data from oVirt API
```
jakub-dzon: RichVmwareClient (...) from VMware API

pkg/providers/vmware/client/client.go
```diff
@@ -0,0 +1,145 @@
...
+
+	vmwareClient := RichVmwareClient{
+		client:         vimClient,
+		url:            u,
+		sessionManager: sessionManager,
+	}
+	return &vmwareClient, nil
+}
+
+func (r RichVmwareClient) GetVM(id *string, _ *string, _ *string, _ *string) (interface{}, error) {
```
jakub-dzon: Please add comments on all exported methods.

pkg/providers/vmware/client/client.go
```diff
@@ -0,0 +1,145 @@
...
+
+// RichOvirtClient is responsible for retrieving VM data from oVirt API
+type RichVmwareClient struct {
+	client         *vim25.Client
+	url            *url.URL
+	sessionManager *session.Manager
+}
+
+// NewRichVMwareClient creates new, connected rich vmware client. After it is no longer needed, call Close().
+func NewRichVMWareClient(apiUrl, username, password string, thumbprint string) (*RichVmwareClient, error) {
```
jakub-dzon, line 28: `Close` method doesn't do anything - "After it is no longer needed, call Close()." part of the comment is not true.

pkg/providers/vmware/client/client.go
```diff
@@ -0,0 +1,145 @@
...
+	vmwareClient := RichVmwareClient{
+		client:         vimClient,
+		url:            u,
+		sessionManager: sessionManager,
+	}
+	return &vmwareClient, nil
+}
+
+func (r RichVmwareClient) GetVM(id *string, _ *string, _ *string, _ *string) (interface{}, error) {
+	return r.getVM(*id), nil
```
jakub-dzon: From the API standpoint VM ID is optional and can be nil -  in that case this line will cause panic, if it is not set to some value in the provider.

pkg/providers/vmware/client/client.go
jakub-dzon: Is name-based lookup unsupported by VMware? If so, please document it with a test case.

pkg/providers/vmware/client/client_test.go
```diff
@@ -0,0 +1,155 @@
...
+	It("should connect to vCenter", func() {
+		_, err := client.NewRichVMWareClient(server.URL.String(), username, password, simulatorThumbprint)
+		Expect(err).To(BeNil())
+	})
+
+	It("should retrieve a VM by ID", func() {
+		richClient, err := client.NewRichVMWareClient(server.URL.String(), username, password, simulatorThumbprint)
+		Expect(err).To(BeNil())
+		vmRef := vm.Reference().Value
+		rawVm, err := richClient.GetVM(&vmRef, nil, nil, nil)
```
jakub-dzon: Please add some "air" to the test cases by separating "given", "when" and "then" parts; for example:

```go
		richClient, err := client.NewRichVMWareClient(server.URL.String(), username, password, simulatorThumbprint)
		Expect(err).To(BeNil())
		vmRef := vm.Reference().Value

		rawVm, err := richClient.GetVM(&vmRef, nil, nil, nil)

		Expect(err).To(BeNil())
		retrievedVm, ok := rawVm.(*object.VirtualMachine)
		Expect(ok).To(BeTrue())
		Expect(retrievedVm.Reference()).To(Equal(vm.Reference()))
```

pkg/providers/vmware/client/client_test.go
```diff
@@ -0,0 +1,155 @@
...
+		err = richClient.StopVM(vmRef)
+		Expect(err).To(BeNil())
+		err = richClient.StartVM(vmRef)
+		Expect(err).To(BeNil())
+		err = richClient.StartVM(vmRef)
+		Expect(err).To(BeNil())
+	})
+})
+
+var _ = Describe("Test VMware rich client with ESXi", func() {
```
jakub-dzon: This set of test cases is exactly the same as for vCenter; it seems to be a good candidate for a table driven test with `model             *simulator.Model` as a parameter.

jakub-dzon (changes requested):

pkg/providers/vmware/client/client_test.go
mansam: Thanks, I was unfamiliar with the concept of table driven tests. Done.

pkg/providers/vmware/client/client.go
mansam: I initially thought that name-based lookup was unsupported, but it appears that finding by inventory path supports bare names. Added.

pkg/providers/vmware/client/client_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	DescribeTable("should retrieve a VM by ID", func(model *simulator.Model) {
+		_ = model.Create()
+		server := model.Service.NewServer()
+		defer model.Remove()
+		defer server.Close()
+		vm := simulator.Map.Any("VirtualMachine").(*simulator.VirtualMachine)
+		username := server.URL.User.Username()
+		password, _ := server.URL.User.Password()
+		richClient, err := client.NewRichVMWareClient(server.URL.String(), username, password, "")
```
jakub-dzon: The `richClient`creation code is the same in most of the test cases; it would be great if you found a way to extract as much of it as possible to a helper function.

pkg/providers/vmware/client/client.go
```diff
@@ -0,0 +1,173 @@
...
+const timeout = 30 * time.Second
+
+// RichVmwareClient is responsible for retrieving VM data from the VMware API.
+type RichVmwareClient struct {
+	client         *vim25.Client
+	user           *url.Userinfo
+	sessionManager *session.Manager
+}
+
+// NewRichVMwareClient creates a new, connected rich VMWare client.
```
jakub-dzon: nit: NewRichVMwareClient ->NewRichVMWareClient in the comment or the other way round with the function name; the linter complains if the first word in the comment on an exported function is not spelled exactly like that function's name.

pkg/providers/vmware/client/client_test.go
```diff
@@ -0,0 +1,137 @@
...
+	)
+
+	DescribeTable("should not throw an error when trying to power on a VM that's already on", func(model *simulator.Model) {
+		_ = model.Create()
+		server := model.Service.NewServer()
+		defer model.Remove()
+		defer server.Close()
+		username := server.URL.User.Username()
+		password, _ := server.URL.User.Password()
+		richClient, err := client.NewRichVMWareClient(server.URL.String(), username, password, "")
```
jakub-dzon: `richClient, err := createRichClient(server)`

jakub-dzon (approved):

pkliczewski (approved):

machacekondra (approved):

ghost:
Fixes https://github.com/kubevirt/vm-import-operator/issues/292

PR #321, opened 2020-07-02
Allow user to specify storage classes for tests 

### LINKED ISSUE #318: Allow user to specify storage-class for tests
https://github.com/kubevirt/vm-import-operator/blob/master/tests/common-vars.go#L5 assumes a storage class named 'local' should be avaiable on the cluster.

It will be usable to expect a custom storage class to be provided and to overrides the default value:
https://github.com/kubevirt/vm-import-operator/blob/master/tests/framework/framework.go#L85

- It is expected that for an external cluster the storage class should be marked as **default** storage-class.
- It is expected to have 'nfs' storage class as well

jakub-dzon:
Fixes #318
```release-notes
NONE
```

### DIFF
--- a/tests/framework/framework.go
+++ b/tests/framework/framework.go
@@ -46,6 +46,8 @@ var (
 	master                   *string
 	ovirtSecretBlueprint     *string
 	kubeVirtInstallNamespace *string
+	defaultStorageClass      *string
+	nfsStorageClass          *string
 )
 
 // This package is based on https://github.com/kubevirt/containerized-data-importer/blob/master/tests/framework/framework.go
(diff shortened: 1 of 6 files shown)


tests/framework/framework.go
```diff
@@ -79,6 +81,11 @@ type Framework struct {
 	OVirtSecretName *types.NamespacedName
 	// KubeVirtInstallNamespace namespace where KubeVirt is installed
 	KubeVirtInstallNamespace string
+
+	// DefaultStorageClass specifies the name of a basic, default storage class
+	DefaultStorageClass string
+	//NfsStorageClass specifies the name of an NFS-based storage class
```
masayag, line 87: nit: add a space before NfsStorageClass

machacekondra (approved):

pkliczewski (approved):

PR #323, opened 2020-07-03
Restore source vm status when import cancelled

pkliczewski:
We need to restore original status of source vm when vm import is cancelled. We use restore finalizer to start source vm if needed.

Bug-Url: https://bugzilla.redhat.com/1853351

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -202,17 +202,12 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 		return reconcile.Result{}, err
 	}
 
-	// Exit if we should not run reconcile:
-	if !shouldReconcile(instance) {
-		reqLogger.Info("Not running reconcile")
-		return reconcile.Result{}, nil
-	}
-
 	// Init provider:
 	provider, err := r.createProvider(instance)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
 	message, err := r.initProvider(instance, provider)
 	if err != nil {
 		if r.vmImportInProgress(instance) {
@@ -226,6 +221,21 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 		return reconcile.Result{}, nil
 	}
 	defer provider.Close()
+
+	if instance.DeletionTimestamp != nil && utils.HasFinalizer(instance, utils.RestoreVMStateFinalizer) {
+		err := r.finalize(instance, provider)
+		if err != nil {
+			// requeue if locked
+			return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
+		}
+		return reconcile.Result{}, nil
+	}
+
+	// Exit if we should not run reconcile:
+	if !shouldReconcile(instance) {
+		reqLogger.Info("Not running reconcile")
+		return reconcile.Result{}, nil
+	}
 
 	// fetch source vm
 	err = r.fetchVM(instance, provider)
@@ -243,7 +253,7 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 	}
 
 	// Stop the VM
-	if err = provider.StopVM(); err != nil {
+	if err = provider.StopVM(instance, r.client); err != nil {
 		return reconcile.Result{}, err
 	}
 
@@ -265,7 +275,7 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
 	}
 
 	// Import disks:
-	err = r.importDisks(provider, instance, mapper, vmName, request.NamespacedName)
+	err = r.importDisks(provider, instance, mapper, vmName)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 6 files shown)


pkg/utils/utils.go
```diff
@@ -247,3 +250,37 @@ func WithLabels(labels map[string]string, existing map[string]string) map[string
 
 	return labels
 }
+
+// AddFinalizer adds finalizer to VM import CR
+func AddFinalizer(cr *v2vv1alpha1.VirtualMachineImport, name string) {
+	if HasFinalizer(cr, name) {
+		return
+	}
+	cr.Finalizers = append(cr.Finalizers, name)
```
machacekondra: Is intentional to not persist the `Finalizers`? It will be changed only to runtime object, and once we re-fetch it, it won't be there.

pkliczewski: Thanks for finding, will make sure it is updated

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -924,9 +944,9 @@ func (r *ReconcileVirtualMachineImport) updateProgress(instance *v2vv1alpha1.Vir
 	return nil
 }
 
-func (r *ReconcileVirtualMachineImport) afterSuccess(vmName types.NamespacedName, vmiName types.NamespacedName, p provider.Provider) error {
+func (r *ReconcileVirtualMachineImport) afterSuccess(vmName types.NamespacedName, vmiName types.NamespacedName, p provider.Provider, instance *v2vv1alpha1.VirtualMachineImport) error {
```
masayag: The same data is passed twice as parameter in a different format:

`instanceNamespacedName := types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}`

The second parameter `vmiName` can be omitted and calculated and be initialized within the method.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -943,9 +963,9 @@ func (r *ReconcileVirtualMachineImport) afterSuccess(vmName types.NamespacedName
 }
 
 //TODO: use in proper places
-func (r *ReconcileVirtualMachineImport) afterFailure(vmiName types.NamespacedName, p provider.Provider) error {
+func (r *ReconcileVirtualMachineImport) afterFailure(vmiName types.NamespacedName, p provider.Provider, instance *v2vv1alpha1.VirtualMachineImport) error {
```
masayag: same comment - we can pass only the new are 'instance' and initialize vmiName within the method:

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -1139,7 +1159,7 @@ func (r *ReconcileVirtualMachineImport) validate(instance *v2vv1alpha1.VirtualMa
 	return true, nil
 }
 
-func (r *ReconcileVirtualMachineImport) templateMatchingFailed(instanceNamespacedName types.NamespacedName, errorMessage string, processingCond *v2vv1alpha1.VirtualMachineImportCondition, provider provider.Provider) error {
+func (r *ReconcileVirtualMachineImport) templateMatchingFailed(instanceNamespacedName types.NamespacedName, errorMessage string, processingCond *v2vv1alpha1.VirtualMachineImportCondition, provider provider.Provider, instance *v2vv1alpha1.VirtualMachineImport) error {
```
masayag: here as well

pkg/utils/utils.go
```diff
@@ -13,11 +14,15 @@ import (
...
 	k8svalidation "k8s.io/apimachinery/pkg/util/validation"
+	rclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
 	// MaxLabelValueLength specifies max length of k8s label value
 	MaxLabelValueLength = 63
+
+	// RestoreVMFinalizer defines restore source vm finalizer
+	RestoreVMFinalizer = "vmimport.v2v.kubevirt.io/restore"
```
masayag: RestoreVM has a different semantic in virt domain, perhaps rename it to RestoreVMStateFinalizer and the annotation name to "vmimport.v2v.kubevirt.io/restore-state"

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -202,17 +202,23 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
 	}
 
+	// Init provider:
+	provider, err := r.createProvider(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if instance.DeletionTimestamp != nil && utils.HasFinalizer(instance, utils.RestoreVMFinalizer) {
+		r.finalize(instance, provider)
```
masayag: We might enter the `r.restoreInitialVMState(vmiName, provider)` in this code path, where the provider wasn't initialized, and as a result to failure to connect to it.

We need to make sure we passed through `r.initProvider` before invoking the `r.finalize`

masayag (changes requested):

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: :+1:

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: OK

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: OK

pkg/utils/utils.go
pkliczewski: :+1:

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: Ack

masayag (approved):

machacekondra (approved):

PR #328, opened 2020-07-07
Add metric kubevirt_vmimport

machacekondra:
Currently tested on `kubernetes` with `kube-prometheus`, the only hack which must be done to make it working is to add `list`, `get` and `watch` verb to resource `services`, `pods` and `endpoints` for `prometheus-k8s` `serviceaccount`. like this:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: prometheus-k8s
rules:
- apiGroups:
  - ""
  resources:
  - nodes/metrics
  - services
  - pods
  - endpoints
  verbs:
  - get
  - list
  - watch
- nonResourceURLs:
  - /metrics
  verbs:
  - get
```

Ready for initial review.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,0 +1,7 @@
+# Virtual Machine Import metrics
+
+Virtual Machine Import uses Prometheus for metrics reporting. The metrics can be used for real-time monitoring. Virtual Machine Import does not persist its metrics, if a member restarts, the metrics will be reset.
+
+| Name               | Description                                                     | Type    | Labels                       |
+|--------------------|-----------------------------------------------------------------|---------|------------------------------|
+| kubevirt_vmimport  | The total number of successfull/failed Virtual Machine imports. | Counter | result=<successful>/<failed> |
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -1,0 +1,47 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const (
+	// MetricsHost host that servers the metrics
+	MetricsHost = "0.0.0.0"
+	// MetricsPort port where metrics are served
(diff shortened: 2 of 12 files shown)


docs/metrics.md
```diff
@@ -0,0 +1,7 @@
+# Virtual Machine Import metrics
+
+Virtual Machine Import uses Prometheus for metrics reporting. The metrics can be used for real-time monitoring. Virtual Machine Import does not persist its metrics, if a member restarts, the metrics will be reset.
+
+| Name                      | Description                                              | Type    |
+|---------------------------|----------------------------------------------------------|---------|
+| kubevirt_vmimport | The total number of Virtual Machine imports.               | Counter |
```
pkliczewski: I think it would be valuable to track vm import result like number how many import failed.

pkliczewski (commented):
The change looks good but I think there is a value in adding metric to track number of failed imports.

docs/metrics.md
machacekondra: Sure, will add.

machacekondra: I've added label `result` with `successful` and `failed`:
Screenshot 2020-07-08 at 13 34 41

pkliczewski (approved):

pkliczewski:
@machacekondra please rebase

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+	// MetricsHost host that servers the metrics
+	MetricsHost = "0.0.0.0"
+	// MetricsPort port where metrics are served
+	MetricsPort int32 = 8383
+	// OperatorMetricsPort port where operator metrics are served
+	OperatorMetricsPort int32 = 8686
+)
+
+var (
+	// ImportCounter counter which hold number of vm imports done
```
masayag: this comment can be removed or fixed to describe importCounterVec

masayag (approved):

PR #331, opened 2020-07-13
Add debug to ovirt client

pkliczewski:
When ovirt client panic we do not have enough information to understand what was the reason of that panic.


masayag (approved):

machacekondra:
The same tests are failing for me as well, nothing changed in our code. So this seems to be related to bump to kubevirt version v0.30.4 from v0.30.3 I will send the change to this version to see whether it's really related.

machacekondra (approved):

PR #335, opened 2020-07-13
Added patchMergeKey and patchStrategy to VirtualMachineImportStatus.Conditions

jakub-dzon:

### DIFF
--- a/docs/functional-tests.md
+++ b/docs/functional-tests.md
@@ -140,10 +140,10 @@
 ## Running tests
 
 ### With kubevirtci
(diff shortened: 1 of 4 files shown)


pkliczewski (commented):
The change looks good. Let's make sure that condition list is not getting overridden.

jakub-dzon:
Thanks! I just wanted to run it against the kubevirt-ci on prow to see if it doesn't break anything. I will be trying to replicate the original problem and only then confirming that the change indeed fixes it.

PR #338, opened 2020-07-14
Fix func tests for kubevirt v0.30.4

machacekondra:
Since kubevirt v0.30.4 secure boot is enabled by default for the EFI
bootloader, so we need to enable SMM, for EFI bootloaders.

### DIFF
--- a/pkg/providers/ovirt/mapper/mapper.go
+++ b/pkg/providers/ovirt/mapper/mapper.go
@@ -40,6 +40,8 @@ const (
 	DefaultStorageClassTargetName = ""
 )
 
+var _true bool = true
+
 // BiosTypeMapping defines mapping of BIOS types between oVirt and kubevirt domains
 var BiosTypeMapping = map[string]*kubevirtv1.Bootloader{
 	"q35_sea_bios":    &kubevirtv1.Bootloader{BIOS: &kubevirtv1.BIOS{}},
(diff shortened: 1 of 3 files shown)


pkliczewski (approved):

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -703,6 +708,23 @@ func (o *OvirtMapper) mapPlacementPolicy() *kubevirtv1.EvictionStrategy {
 	return nil
 }
 
+func (o *OvirtMapper) mapFeatures() *kubevirtv1.Features {
```
jakub-dzon, line 711: Please add proper check in the unit test.

jakub-dzon (changes requested):

pkg/providers/ovirt/mapper/mapper_test.go
```diff
@@ -29,6 +29,43 @@ var (
...
+			URL:           "any-url",
+			SecretName:    "secret-name",
+			ConfigMapName: "config-map",
+		}
+		namespace := "the-namespace"
+		mapper := mapper.NewOvirtMapper(vm, &mappings, credentials, namespace, &osFinder)
+		vmSpec, _ := mapper.MapVM(&targetVMName, &kubevirtv1.VirtualMachine{})
+
+		if *smm {
+			Expect(vmSpec.Spec.Template.Spec.Domain.Features).ToNot(BeNil())
```
jakub-dzon: That check can be moved before if - the other branch assumes that `vmSpec.Spec.Template.Spec.Domain.Features` are not nil.

jakub-dzon (approved):

PR #339, opened 2020-07-14
Release v0.1.0

jakub-dzon:
Releasing v0.0.4 as v0.1.0 into release-0.1 line branch.
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 all: fmt check
 
 # Always keep the last released version here
-VERSION_REPLACES ?= v0.0.4
+VERSION_REPLACES ?= v0.1.0
 VERSION ?= v0.0.1
 export VERSION := $(VERSION)
 
--- a/README.md
+++ b/README.md
@@ -17,8 +17,8 @@ TARGET_NAMESPACE=YOUR_DESIRED_NAMESPACE make gen-manifests
 
 Deploy vm-import-operator resources:
 ```bash
-kubectl apply -f https://github.com/kubevirt/vm-import-operator/releases/download/v0.0.4/operator.yaml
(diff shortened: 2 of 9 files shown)


pkliczewski (approved):

pkliczewski:
@jakub-dzon I am not sure why we should have kubevirt version bump as part of the stable branch.

masayag:
> @jakub-dzon I am not sure why we should have kubevirt version bump as part of the stable branch.

I'd also prefer to keep the patch only about bumping the version and not include other changes as part of it.

jakub-dzon:
Without that change the functional tests will fail, as we always take the newest release, and v0.30.4 proved to introduce changes that break our functionality. I wanted to avoid having a release PR with failed functional tests.

masayag (approved):

masayag:
> Without that change the functional tests will fail, as we always take the newest release, and v0.30.4 proved to introduce changes that break our functionality. I wanted to avoid having a release PR with failed functional tests.

Thanks for the explanation. In that case we'll have to revert the changes related to CDI and kubevirtci versions. Should an issue track this ? Also, are there known issues that can be reported as bugs to CDI/kubevirt ?

jakub-dzon:
@masayag: Ondra has already implemented proper fix on the master branch: https://github.com/kubevirt/vm-import-operator/pull/338; the PR also explains what was the issue. I didn't want to pull that in, as it changes the production code. IMO we should backport it and ship in v0.1.1.

pkliczewski:
@machacekondra what do you think?

machacekondra:
I agree with Jakub. Let's release the fix with another version and let's bump the kubevirt & CDI versions here.

machacekondra (approved):

PR #346, opened 2020-07-17
Using JSON Patch for updating conditions

### LINKED ISSUE #322: VMIMports condition may override existing condition
When running a VM Import that fails due to a missing secret (or other failure to initialize the provider),  it may be that both conditions are recorded as expected, e.g. for the expected scenario the status should contain both condition types 'Valid' and 'Succeedeed' :
```
  conditions:
  - lastHeartbeatTime: "2020-07-02T16:20:16Z"
    lastTransitionTime: "2020-07-02T16:20:16Z"
    message: 'Failed to connect to source provider: Post /ovirt-engine/sso/oauth/token:
      unsupported protocol scheme ""'
    reason: UnreachableProvider
    status: "False"
    type: Valid
  - lastHeartbeatTime: "2020-07-02T16:20:16Z"
    lastTransitionTime: "2020-07-02T16:20:16Z"
    message: 'Failed to initialize the source provider (Failed to connect to source
      provider): Post /ovirt-engine/sso/oauth/token: unsupported protocol scheme ""'
    reason: ValidationFailed
    status: "False"
    type: Succeeded
```

However in some cases only the latter condition is kept and overrides the previous condition.
When watching the status of the CR it looks that 'Valid' is recorded:
```
status:
  conditions:
  - lastHeartbeatTime: "2020-07-02T15:12:41Z"
    lastTransitionTime: "2020-07-02T15:12:41Z"
    message: Secret not found
    reason: SecretNotFound
    status: "False"
    type: Valid
```
And then it is overridden with:
```
status:
  conditions:
  - lastHeartbeatTime: "2020-07-02T15:12:41Z"
    lastTransitionTime: "2020-07-02T15:12:41Z"
    message: 'Failed to initialize the source provider (Secret not found): Secret
      "no-such-secret" not found'
    reason: ValidationFailed
    status: "False"
    type: Succeeded
```

The expected behavior is to maintain all of the condition that occurred during the run of the vm-import process.

Noticed on https://github.com/kubevirt/vm-import-operator/pull/310#discussion_r449201548
When this issue is fixed, tests can be changed to expect the 'Valid' condition with the desired reason for the validation failure.

jakub-dzon:
Fixes #322 by using VM Import CRD retrieved directly from the cluster as a base for changes and status update for persisting the changes.

### DIFF
--- a/pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
+++ b/pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
@@ -90,7 +90,7 @@ type VirtualMachineImportStatus struct {
 	Conditions []VirtualMachineImportCondition `json:"conditions"`
 
 	// +optional
-	DataVolumes []DataVolumeItem `json:"dataVolumes"`
+	DataVolumes []DataVolumeItem `json:"dataVolumes,omitempty"`
 }
 
 // VirtualMachineImportConditionType defines the condition of VM import
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -103,11 +103,13 @@ func newReconciler(mgr manager.Manager, kvConfigProvider config.KubeVirtConfigPr
 		log.Error(err, "Unable to get OC client")
 		panic("Controller cannot operate without OC client")
 	}
+	reader := mgr.GetAPIReader()
 	client := mgr.GetClient()
 	finder := mappings.NewResourceMappingsFinder(client)
 	ownerreferencesmgr := ownerreferences.NewOwnerReferenceManager(client)
 	factory := pclient.NewSourceClientFactory()
 	return &ReconcileVirtualMachineImport{client: client,
+		apiReader:              reader,
 		scheme:                 mgr.GetScheme(),
 		resourceMappingsFinder: finder,
 		ocClient:               tempClient,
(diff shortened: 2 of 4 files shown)


pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/appscode/jsonpatch"
```
pkliczewski: This repo seems not to be actively maintained. Do you see us maintaining it in the future?

jakub-dzon: Good point. So far I'm checking if the general approach fixes the problem; might switch to something else.

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
```diff
@@ -90,7 +90,7 @@ type VirtualMachineImportStatus struct {
 	Conditions []VirtualMachineImportCondition `json:"conditions"`
 
 	// +optional
-	DataVolumes []DataVolumeItem `json:"dataVolumes"`
+	DataVolumes []DataVolumeItem `json:"dataVolumes,omitempty"`
```
pkliczewski, line 93: Is this relevant to updating conditions?

machacekondra (approved):
Did you notice any performance issue, when using API directly instead of cache?

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
jakub-dzon: Kind of; without it the update on status without this slice set fails API-level validation - at the early stages of import processing GET returns it set to nil.

jakub-dzon:
@machacekondra - I didn't see any issues related to that. It doesn't seem to take the import more time to get to the desired state - at least functional tests don't take longer than they used to.

pkliczewski (approved):

PR #347, opened 2020-07-21
Add support for block volumes

machacekondra:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1857926


pkg/operator/resources/operator/operator.go
```diff
@@ -726,6 +726,9 @@ attachment definitions on kubevirt by using vnic profile to network attachment d
 																		"type": {
 																			Type: "string",
 																		},
+																		"volumeMode": {
```
jakub-dzon, line 729: The description of the object needs to be fixed; it refers to `ResourceMappingItem` which no longer exist ("ResourceMappingItem defines the...") (line 699).

pkg/operator/resources/operator/operator.go
```diff
@@ -769,6 +772,9 @@ the disk alias on ovirt DiskMappings is respected only when provided in context
...
```
jakub-dzon, line 775: Ditto

pkg/operator/resources/operator/operator.go
```diff
@@ -1071,6 +1080,9 @@ the disk alias on ovirt DiskMappings is respected only when provided in context
 														"type": {
 															Type: "string",
 														},
+														"volumeMode": {
```
jakub-dzon, line 1083: Ditto

pkg/operator/resources/operator/operator.go
```diff
@@ -1028,6 +1034,9 @@ to multiple network attachment definitions on kubevirt by using vnic profile to
...
```
jakub-dzon, line 1037: Ditto

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -261,6 +301,7 @@ func (o *OvirtMapper) MapDataVolumes(targetVMName *string) (map[string]cdiv1.Dat
 		}
 		quantity, _ := resource.ParseQuantity(diskSizeConverted)
 		accessMode := o.getAccessMode(diskAttachment)
+		volumeMode := o.getVolumeMode(disk, o.mappings)
```
jakub-dzon: To detect the volume mode we're iterating over the same mappings slice for the second time (first - for the storage class in `getStorageClassForDisk`). In both cases we should get exactly the same mapping instance as a base for retrieving both storage class and volume mode. What do you think about finding the mapping first and than passing it to `getVolumeMode` and `getStorageClassForDisk` to extract the needed parameter?

pkg/providers/ovirt/mappings/merger.go
```diff
@@ -38,16 +38,62 @@ func extractMappings(externalMappingSpec *v2vv1alpha1.ResourceMappingSpec, crMap
 	return &primaryMappings, &secondaryMappings
 }
 
-func mergeMappings(primaryMappings *[]v2vv1alpha1.ResourceMappingItem, secondaryMappings *[]v2vv1alpha1.ResourceMappingItem) *[]v2vv1alpha1.ResourceMappingItem {
-	var mapping []v2vv1alpha1.ResourceMappingItem
+func mergeNetworkMappings(primaryMappings *[]v2vv1alpha1.NetworkResourceMappingItem, secondaryMappings *[]v2vv1alpha1.NetworkResourceMappingItem) *[]v2vv1alpha1.NetworkResourceMappingItem {
```
jakub-dzon, line 41: I would prefer both `mergeNetworkMappings` and `mergeStorageMappings` to use the same common code, but without common interface or generics it's unlikely that it can be done cleanly. Definitely a good candidate for trying them when they're finally supported.;)

pkg/providers/ovirt/mapper/mapper_test.go
```diff
@@ -370,7 +370,7 @@ var _ = Describe("Test mapping disks", func() {
 	It("should map empty disk storage class to nil", func() {
 		diskID := "disk-ID"
 		targetStorageClass := ""
-		disks := []v2vv1alpha1.ResourceMappingItem{
+		disks := []v2vv1alpha1.StorageResourceMappingItem{
```
jakub-dzon, line 377: Please add test/check for the volume mode.

jakub-dzon (changes requested):

pkg/providers/ovirt/mappings/merger.go
machacekondra: Yeah, I was trying to solve it with reflection, but it was even worse IMHO, so I left the code duplication.

pkg/providers/ovirt/mapper/mapper.go
machacekondra: Good point, thanks.

jakub-dzon (approved):

pkliczewski (approved):

pkliczewski:
@machacekondra I see 2 conflicting files. Would you mind to rebase?

machacekondra:
> @machacekondra I see 2 conflicting files. Would you mind to rebase?

Done.

PR #350, opened 2020-07-24
Check if CRD have stuctural schema

machacekondra:

### DIFF
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -31,7 +31,14 @@ fi
 
 ensure_golang
 
+## Validate
+operator_crds=$(./cluster/kubectl.sh get crd -l operator.v2v.kubevirt.io -o jsonpath={.items[*].metadata.name})
+check_structural_schema "${operator_crds[@]}"
+
+## Configure
 configure_nfs
+
+## Install
 install_cdi
 install_kubevirt
 install_templates
--- a/cluster/common.sh
+++ b/cluster/common.sh
@@ -66,3 +66,15 @@ spec:
 EOF
 
 }
+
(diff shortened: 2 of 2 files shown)


cluster/up.sh
```diff
@@ -31,7 +31,14 @@ fi
 
 ensure_golang
 
+## Validate
+operator_crds=$(./cluster/kubectl.sh get crd -l operator.v2v.kubevirt.io -o jsonpath={.items[*].metadata.name})
+check_structural_schema "${operator_crds[@]}"
```
jakub-dzon, line 36: When was the `NonStructuralSchema` condition reported as `True`? What should be investigated when this validation fails?

machacekondra: The k8s report that in case the CRD is not structural. In that case the CRD API is wrong, there are rules, which must comply to have structural schema: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

jakub-dzon: So it looks like it boils down to fixing something in https://github.com/kubevirt/vm-import-operator/blob/master/pkg/operator/resources/operator/operator.go. Is that correct?

jakub-dzon (approved):

cluster/up.sh
machacekondra: Right, in one of `resourcemappings.v2v.kubevirt.io`, `virtualmachineimports.v2v.kubevirt.io`, `vmimportconfigs.v2v.kubevirt.io`.

pkliczewski (approved):

PR #354, opened 2020-07-29
Add Valid condition warning when default storage class is about to be used

jakub-dzon:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1853403

When the default storage class is going to be used for a disk, `Valid` condition will look like the one below:
```yaml
    - lastHeartbeatTime: "2020-07-29T12:08:45Z"
      lastTransitionTime: "2020-07-29T12:08:45Z"
      message: Default storage class will be used for disk disk-1(disk-1)
      reason: ValidationReportedWarnings
      status: "True"
      type: Valid
```
```release-note
NONE
```

### DIFF
--- a/pkg/apis/v2v/v1beta1/virtualmachineimport_types.go
+++ b/pkg/apis/v2v/v1beta1/virtualmachineimport_types.go
@@ -163,6 +163,9 @@ const (
 
 	// IncompleteMappingRules represents the inability to prepare the mapping rules
 	IncompleteMappingRules ValidConditionReason = "IncompleteMappingRules"
+
+	// ValidationReportedWarnings represents the existence of warnings related to resource mapping validation
+	ValidationReportedWarnings ValidConditionReason = "ValidationReportedWarnings"
 )
 
 // MappingRulesVerifiedReason defines the reasons for the MappingRulesVerified condition of VM import
--- a/pkg/providers/ovirt/validation/validators/storage-mapping-validator.go
+++ b/pkg/providers/ovirt/validation/validators/storage-mapping-validator.go
@@ -36,6 +36,10 @@ const (
 	storageDomainSourceType = sourceType("storage domain")
 )
 
+var (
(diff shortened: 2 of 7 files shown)


pkg/apis/v2v/v1beta1/virtualmachineimport_types.go
```diff
@@ -163,6 +163,9 @@ const (
 
 	// IncompleteMappingRules represents the inability to prepare the mapping rules
 	IncompleteMappingRules ValidConditionReason = "IncompleteMappingRules"
+
+	// ValidationReportedWarnings represents the existence of warnings as a result of checking the mapping
```
pkliczewski: Would it be possible to rephrase `represents the existence of warnings as a result of checking the mapping`. It took me some time to understand what it means.

pkg/providers/ovirt/validation/validators/storage-mapping-validator.go
```diff
@@ -112,6 +115,13 @@ func (v *StorageMappingValidator) validateStorageClasses(requiredTargetsSet map[
 	var failures []ValidationFailure
 	for className, sources := range requiredTargetsSet {
 		if className.Name == mapper.DefaultStorageClassTargetName {
+			for _, source := range sources {
+				resourceName := utils.ToLoggableID(source.ID, source.Name)
+				failures = append(failures, ValidationFailure{
+					ID:      StorageTargetDefaultClass,
+					Message: fmt.Sprintf("Default storage class will be used for disk %s", resourceName),
```
pkliczewski: What do you think to change message to: `be used for %s disk`?

pkg/apis/v2v/v1beta1/virtualmachineimport_types.go
jakub-dzon: What about "ValidationReportedWarnings represents the existence of warnings related to resource mapping validation"?

pkliczewski: Sounds good

pkliczewski (approved):

PR #358, opened 2020-07-31
Controller config map backed os mapping

### LINKED ISSUE #356: Allow user to update OS mapping ConfigMap location without triggerring redeploy
To address https://bugzilla.redhat.com/1847810:

New ConfigMap will be introduced to allow users to customize the runtime behaviour of the vm-import-controller, including the namespace and name of the OS mapping ConfigMap. Changes made to the entries in the new config map will be used by the vm-import-controller immediately and without restart. 

The new ConfigMap will reside in the same namespace as the vm-import-controller and will be created by the vm-import-operator if it does not exist. Configuration currently defined on the vm-import-operator level (env variables) will be migrated to the new approach by the operator when the new ConfigMap is created.   
 
The structure of the OS mapping config map will not be changed.
 
The contents of the config map must be stored in must-gather.
 
Design documents and related documentation needs to be updated.

jakub-dzon:
Fixes #356 
This PR introduces use of config map properties instead of environment variables to provide the location of os mapping config map.

```release-notes
NONE
```


cluster/operator-install.sh
```diff
@@ -20,6 +20,8 @@ export VERSION=v0.0.0
 export VM_IMPORT_CONFIG_STATUS=${VM_IMPORT_CONFIG_STATUS:-''}
 
 # Create and wait for the operator
+cat tests/os-mapping/operator.yaml.patch >> _out/vm-import-operator/${VERSION}/operator.yaml
```
masayag, line 23: what is this for ?

docs/design.md
```diff
@@ -172,55 +172,41 @@ kind: ConfigMap
...
+kind: ConfigMap
+metadata:
+  name: vm-import-controller-config
+  namespace: kubevirt
+```
 
-In order to update `vm-import-operator` deployment, user may use CLI or UI for setting environment variable values, e.g.:
-![Image of vm-import-operator deployment](/docs/images/vm-import-operator-deployment.png)
+In order to update `vm-import-controller-config` deployment, user may use CLI or UI for setting the configuration values, e.g.:
+![Image of vm-import-controller-config config map](/docs/images/controller_config_map.png)
```
masayag, line 194: I'd clear unneeded fields from the attached image, e.g. self-link, uid, resource-version, creation timestamp - in order to emphasis only the required or important fields in this context

cluster/operator-install.sh
jakub-dzon: That's for the migration from the environment variables functional test - it adds the OS map-related  env variables to the vm-import-operator deployment (which are no longer there) so it is possible to verify that they were migrated to the config map - in https://github.com/kubevirt/vm-import-operator/blob/a7785a22a286f2ce863ee349534a9927e929c037/tests/ovirt/user_provided_os_maps_migration_test.go.

pkg/config/controller/config-provider.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 const (
-	configMapName = "vm-import-controller-config"
+	ConfigMapName = "vm-import-controller-config"
```
masayag, line 10: that requires a comment ? (due to init-cap)

pkg/config/controller/config.go
```diff
@@ -3,9 +3,9 @@ package controller
 import "github.com/kubevirt/vm-import-operator/pkg/config"
 
 const (
-	osConfigMapNamespaceKey = "osConfigMap.namespace"
+	OsConfigMapNamespaceKey = "osConfigMap.namespace"
```
masayag, line 7: would it require comment since it is exported ?

pkg/operator/controller/controller.go
```diff
@@ -934,3 +929,45 @@ func (r *ReconcileVMImportConfig) isMutable(obj runtime.Object) bool {
...
+	}
+	return nil
+}
+
+func (r *ReconcileVMImportConfig) getOsMappingConfigMapName(namespace string) *types.NamespacedName {
+	var configMapName, configMapNamespace string
+	operatorDeployment := &appsv1.Deployment{}
+	key := client.ObjectKey{Namespace: namespace, Name: "vm-import-operator"}
+	if err := r.client.Get(context.TODO(), key, operatorDeployment); err == nil {
+		operatorEnv := operatorDeployment.Spec.Template.Spec.Containers[0].Env
```
masayag: just making sure - there is expected to be only single container per operator's deployment or would it be safer relying on the container-name ?

pkg/operator/controller/controller.go
```diff
@@ -934,3 +929,45 @@ func (r *ReconcileVMImportConfig) isMutable(obj runtime.Object) bool {
...
+		for _, env := range operatorEnv {
```
masayag, line 969: Can you add an event  / warn that the deployment contains env-vars which are no longer supported and it was replaced by the config-map ?
We probably can't edit or remove the env variables from an existing deployment, but we should somehow let the user be aware that the configuration contains an obsolete vars (assuming the vm-import-operator is installed by the HCO CSV).

pkg/os/os-mapper-provider.go
```diff
@@ -53,22 +54,20 @@ func (o *OSMaps) GetOSMaps() (map[string]string, map[string]string, error) {
...
 	osConfigMap := &corev1.ConfigMap{}
-	osConfigMapName := os.Getenv(OsConfigMapName)
-	osConfigMapNamespace := os.Getenv(OsConfigMapNamespace)
-	if osConfigMapName == "" && osConfigMapNamespace == "" {
+	if o.osConfigMapName == "" && o.osConfigMapNamespace == "" {
 		return nil
 	}
-
-	err := o.Client.Get(
+	log.Info("Using user-provided OS maps", "name", o.osConfigMapName, "namespace", o.osConfigMapNamespace)
```
masayag, line 60: just a note - this is a generic message not tied to any particular vm-import request.
but it seems enough to me, since we don't expect too frequent changes to it.

pkg/operator/controller/controller.go
jakub-dzon: You're right - it will be safer to use the name.

pkg/operator/controller/controller.go
jakub-dzon: Fixed.

pkg/operator/controller/controller.go
jakub-dzon: Fixed.

pkg/config/controller/config.go
jakub-dzon: Fixed.

pkg/config/controller/config-provider.go
jakub-dzon: Fixed.

docs/design.md
jakub-dzon: Fixed.

masayag (approved):

pkliczewski (approved):

PR #360, opened 2020-08-05
Add pending condition

machacekondra:

### DIFF
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -373,6 +373,7 @@ func (r *ReconcileVirtualMachineImport) importDisks(provider provider.Provider, 
 				}
 			}
 		} else if err == nil {
+			instanceNamespacedName := types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}
 			// Set dataVolume as done, if it's in Succeeded state:
 			if foundDv.Status.Phase == cdiv1.Succeeded {
 				dvsDone[dvID] = true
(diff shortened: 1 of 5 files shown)


pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -380,6 +380,13 @@ func (r *ReconcileVirtualMachineImport) importDisks(provider provider.Provider,
 				if err = r.endImportAsFailed(provider, instance, foundDv, "dv is in Failed Phase"); err != nil {
 					return err
 				}
+			} else if foundDv.Status.Phase == cdiv1.Pending {
+				instanceNamespacedName := types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}
+				message := fmt.Sprintf("DataVolume %s is pending to bound", foundDv.Name)
+				processingCond := conditions.NewProcessingCondition(string(v2vv1.Pending), message, corev1.ConditionTrue)
```
pkliczewski, line 387: Do we want to remove this condition once the DV got bound?

machacekondra: Right, thanks, it may be there for long time. Do we want to create another condition like 'Bound', or switching condition back to 'CopyingDisks' cond is enough?

machacekondra: IMHO, 'CopyingDisks' is better.

machacekondra:
```
- lastHeartbeatTime: "2020-08-05T12:53:50Z"
      lastTransitionTime: "2020-08-05T12:53:40Z"
      message: DataVolume testvm2-26097887-1f4d-4718-961f-f5b63a49c3f5 is pending
        to bound
      reason: Pending
      status: "True"
      type: Processing
```

pkliczewski (approved):

PR #366, opened 2020-08-10
Add vSphere VM mapper

mansam:
* Adds the OSFinder, TemplateFinder, and Mapper for the vSphere provider.
* Updates the guest OS identifier mappings to include vSphere guest identifiers
* Refactors some of the mapping merger code to be shared between vSphere and oVirt
* Adds prerecorded vcsim data for testing


jakub-dzon:
@mansam - could you please break this PR down into smaller, independent ones? Smaller PRs will be more inviting and easier to comprehend.

PR #368, opened 2020-08-11
Select newest template

### LINKED ISSUE #345: Select latest template
When selecting a template for a source vm we should make sure to use the latest (based on the crationTimestamp).

pkliczewski:
Whenever we select templates we may get more than one for the same labels with this PR we make sure to always used the latest based on template creation timestamp.

Fixes: https://github.com/kubevirt/vm-import-operator/issues/345

### DIFF
--- a/pkg/providers/ovirt/templates/template-finder_test.go
+++ b/pkg/providers/ovirt/templates/template-finder_test.go
@@ -2,6 +2,7 @@ package templates_test
 
 import (
 	"fmt"
+	"time"
 
 	otemplates "github.com/kubevirt/vm-import-operator/pkg/providers/ovirt/templates"
 	"github.com/kubevirt/vm-import-operator/pkg/templates"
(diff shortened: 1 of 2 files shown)


pkg/providers/ovirt/templates/template-finder_test.go
```diff
@@ -63,6 +64,28 @@ var _ = Describe("Finding a Template", func() {
...
+			return templateList, nil
+		}
+
+		vmOS := ovirtsdk.OperatingSystem{}
+		vmOS.SetType("rhel")
+		vm := ovirtsdk.NewVmBuilder().Os(&vmOS).MustBuild()
+		template, err := templateFinder.FindTemplate(vm)
+
+		Expect(err).To(BeNil())
+		Expect(newer.Equal(&template.CreationTimestamp)).To(BeTrue())
```
jakub-dzon: Is there a reason for this assertion not to be
```go
Expect(template.CreationTimestamp).To(Equal(newer))
```
?

pkliczewski: This `time` type implements its own `Equal` method. You can find it [here](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/time.go#L77)

jakub-dzon: I may be nitpicking but to me it looks rather unusual and by it suggests that some extra measures need to be applied to verify the behaviour. 

Are the special features of `time.Time.Equal` really needed? As per documentation it allows for two "times" to be equal if they are in different locations but represent the same instant - likely in contrast to the `reflect.DeepEqual` that is used by the matcher. Do we, from the client perspective, allow the code under test to change the location of the time?

pkliczewski: I think one of the benefit for using k8s type `Equal` is handling `nil`. For this specific test we know both timestamps won't be `nil` but it feels right to use `Equal` defined for the type.

pkliczewski: If you feel we should use standard matcher I am ok with it.

jakub-dzon: The reasoning behind using the dedicated `Equal` absolutely makes sense. My reluctance towards using it is caused by the sheer surprise of seeing an uncommon assertion construction that made me stop and investigate why it looks this way. When the code changes in the future someone will probably need to investigate it again.
I don't want to push it, though.

jakub-dzon: Thanks for the change:)

jakub-dzon (approved):

pkg/providers/ovirt/templates/template-finder_test.go
pkliczewski: Too late. I have updated the PR :stuck_out_tongue:

machacekondra (approved):

PR #376, opened 2020-08-13
Add virt-v2v container image

ghost:
This pull request adds the assets to build the container image used to convert a VMware VM. It has two files:

- `Dockerfile` used to build the image. It installs all the dependencies (`virt-v2v` and `qemu-img`) and the entrypoint script
- `entrypoint` that scripts the usage of `virt-v2v` and `qemu-img` to do an _in-place_ conversion.

```release-note
Adds assets to build the container image that runs virt-v2v
```


ghost:
@mrnold, I directly proposed the image under vm-import-operator, as it's not related to CDI. Hope you don't mind :smiley:

build/virt-v2v/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM centos:8
```
pkliczewski, line 1: What do you think about using `registry.access.redhat.com/ubi8/ubi-minimal`. I see that good number of containers in `kubevirt` org is using this base image.

pkliczewski (approved):

build/virt-v2v/Dockerfile
ghost: Unfortunately, `ubi8` is not able to pull `qemu-guest-agent` package, and others. The `ubi-8-appstream` mainly contains packages related to development, like compilers, libraries..., so nothing related to more common OS usage.

PR #378, opened 2020-08-19
Update types

pkliczewski:
We need to keep CRD version history as well we were missing status.phase and condition.type.

Bug-Url: https://bugzilla.redhat.com/1868229

### DIFF
--- a/pkg/operator/resources/operator/operator.go
+++ b/pkg/operator/resources/operator/operator.go
@@ -426,9 +426,9 @@ func CreateVMImportConfig() *extv1.CustomResourceDefinition {
 			Scope: "Cluster",
 			Versions: []extv1.CustomResourceDefinitionVersion{
 				{
-					Name:    "v1beta1",
+					Name:    "v1alpha1",
 					Served:  true,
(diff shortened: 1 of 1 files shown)


pkliczewski:
test failure not related to this PR

jakub-dzon (approved):

pkg/operator/resources/operator/operator.go
```diff
@@ -1373,6 +1479,186 @@ func CreateResourceMapping() *extv1.CustomResourceDefinition {
...
+																		"namespace": {
+																			Type: "string",
+																		},
+																	},
+																	Required: []string{"name"},
+																},
+																"type": {
+																	Type: "string",
+																},
+																"volumeMode": {
```
machacekondra: This should be part only as part of v1beta1 API, please use v1alpha1 schema before this change was merged.

pkliczewski: Good point, my copy of `v1alpha1` was from just before your changes. Will fix, Thanks!

machacekondra (approved):

PR #381, opened 2020-08-20
Add vSphere VM mapper

mansam:
This adds a VM mapper for the vSphere provider. Most of the files in this PR are simulator data used for the tests. This is using a Blank DataVolume source as a placeholder until the VDDK datasource is merged.


pkg/providers/vmware/mapper/mapper.go
```diff
@@ -0,0 +1,562 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubevirtv1 "kubevirt.io/client-go/api/v1"
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
+)
+
+const (
+	cdiAPIVersion                 = "cdi.kubevirt.io/v1alpha1"
+	dataVolumeKind                = "DataVolume"
+	DefaultStorageClassTargetName = ""
```
jakub-dzon: Please make it unexported; it doesn't seem to be used anywhere outside of this package.

pkg/providers/vmware/mapper/mapper.go
```diff
@@ -0,0 +1,562 @@
...
+	LabelTag                      = "tags"
```
jakub-dzon: Ditto

pkg/providers/vmware/mapper/mapper.go
```diff
@@ -0,0 +1,562 @@
...
+	bus             string
+	capacity        resource.Quantity
+	datastore       string
+	id              string
+	name            string
+}
+
+// nic is an abstraction of a VMWare VirtualEthernetCard
+type nic struct {
+	name string
```
jakub-dzon: Please run `go fmt` on this file.

pkg/providers/vmware/mapper/mapper.go
```diff
@@ -0,0 +1,562 @@
...
+				}
+			}
+		}
+	}
+
+	if r.mappings.StorageMappings != nil {
+		for _, mapping := range *r.mappings.StorageMappings {
+			// compare datastore moRef
+			if disk.datastore == *mapping.Source.ID {
+				return &mapping.Target.Name
```
jakub-dzon: Shouldn't it be checked against the `DefaultStorageClassTargetName` here as well?

pkg/providers/vmware/mapper/mapper.go
```diff
@@ -0,0 +1,562 @@
...
+		}
+		r.disksByDVName[dvName] = disk
+	}
+	return dvs, nil
+}
+
+// MapDisk maps a disk from the VMware VM to the Kubevirt VM.
+func (r *VmwareMapper) MapDisk(vmSpec *kubevirtv1.VirtualMachine, dv cdiv1.DataVolume) {
+	volume := kubevirtv1.Volume{
+		Name: fmt.Sprintf("dv-%v", dv.Name),
```
jakub-dzon: The name cannot be longer than 63 characters (due to https://bugzilla.redhat.com/show_bug.cgi?id=1857165#c2).
oVirt mapper uses `name = utils.EnsureLabelValueLength(name)` to make sure it's OK.

pkg/providers/vmware/mapper/mapper_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package mapper_test
```
jakub-dzon, line 1: Please run `go fmt` on this file.

jakub-dzon (changes requested):

pkg/providers/vmware/mapper/mapper.go
```diff
@@ -0,0 +1,567 @@
...
+		namespace:      namespace,
+		osFinder:       osFinder,
+		vm:             vm,
+		vmProperties:   vmProperties,
+	}
+}
+
+// buildDevices retrieves each of the VM's VirtualDisks and VirtualEthernetCards
+// and pulls out the values that are needed for import
+func (r *VmwareMapper) buildDevices() error {
```
pkliczewski: Why not to have 2 different methods? one would be to build `disks` and the other `nics`.

pkg/providers/vmware/mapper/mapper.go
```diff
@@ -0,0 +1,567 @@
...
+			}
+
+			disks = append(disks, disk)
+			continue
+		}
+
+		// is this device a VirtualEthernetCard?
+		var virtualNetwork *types.VirtualEthernetCard
+		switch v := device.(type) {
+		case *types.VirtualE1000:
```
pkliczewski: we could simplify this code by using multiple case line.

pkg/providers/vmware/mapper/mapper.go
```diff
@@ -0,0 +1,567 @@
...
+				APIVersion: cdiAPIVersion,
+				Kind:       dataVolumeKind,
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      dvName,
+				Namespace: r.namespace,
+			},
+			Spec: cdiv1.DataVolumeSpec{
+				Source: cdiv1.DataVolumeSource{
+					Blank: &cdiv1.DataVolumeBlankImage{},
```
pkliczewski, line 256: Will the `Source` be update in the next PR?

pkg/providers/vmware/mapper/mapper.go
```diff
@@ -0,0 +1,567 @@
...
+			},
+		},
+	}
+
+	disk := r.disksByDVName[dv.Name]
+	kubevirtDisk := kubevirtv1.Disk{
+		Name: name,
+		DiskDevice: kubevirtv1.DiskDevice{
+			Disk: &kubevirtv1.DiskTarget{
+				Bus: disk.bus,
```
pkliczewski: It seems like we always use `busTypeVirtio`. Are there any plans to support more?

pkg/providers/vmware/mapper/mapper.go
mansam: Mostly I was trying to avoid iterating over that array twice, but I think you're right that it would be better to split it up for organizational purposes.

pkg/providers/vmware/mapper/mapper.go
mansam: Once https://github.com/kubevirt/containerized-data-importer/pull/1320 is merged I'll be making a follow up to set the correct DataVolumeSource.

pkg/providers/vmware/mapper/mapper.go
mansam: The virt-v2v conversion installs the virtio drivers on the guest, so they should be available for all devices. I'm unfamiliar with the implementation details here: Is there a benefit to using scsi/sata if virtio is available? @fdupont-redhat thoughts?

ghost: No benefit. You're correct: virt-v2v installs virtio drivers (that's the bus type in libvirt XML input), so the operating system will be able to use them. The only thing I'm wondering is if CNV is using virtio-blk or virtio-scsi when setting `busTypeVirtio`. @pkliczewski do you know? In RHV, we preferred virtio-scsi as it supports more features and is supposedly more performant.

pkliczewski: My point was that we have a `bus` field in `disk` struct and we always use virtio. I asked the question and later wanted to suggest just to use `busTypeVirtio` here and remove it from the struct.

mansam: Ah yes, will do.

pkg/providers/vmware/mapper/mapper.go
mansam: Putting the types in a case list unfortunately does not work. `v` gets treated as `BaseVirtualDevice` in the combined statement, and does not have the `.VirtualEthernetCard` property.

jakub-dzon (approved):

pkliczewski (approved):

PR #382, opened 2020-08-24
Add provider to import_counter metric

machacekondra:

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -304,3 +304,14 @@ func RemoveFinalizer(cr *v2vv1.VirtualMachineImport, name string, client rclient
 	patch := rclient.MergeFrom(cr)
 	return client.Patch(context.TODO(), copy, patch)
 }
+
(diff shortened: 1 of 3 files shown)


pkg/utils/utils.go
```diff
@@ -304,3 +304,14 @@ func RemoveFinalizer(cr *v2vv1.VirtualMachineImport, name string, client rclient
 	patch := rclient.MergeFrom(cr)
 	return client.Patch(context.TODO(), copy, patch)
 }
+
+// GetProviderName return name of the source provider
+func GetProviderName(cr *v2vv1.VirtualMachineImport) string {
+	if cr.Spec.Source.Ovirt != nil {
+		return "oVirt"
+	} else if cr.Spec.Source.Ovirt != nil {
```
masayag: shouldn't it be vmware ?

machacekondra: Right, thanks.

PR #388, opened 2020-08-27
Add VmwareProvider

mansam:
This implements the Provider interface for the Vmware package, and it makes Vmware one of the supported source types in `createProvider`. The next PR will add the guest conversion job to the import controller.


pkg/providers/vmware/mapper/mapper.go
```diff
@@ -298,7 +299,11 @@ func (r *VmwareMapper) MapDisk(vmSpec *kubevirtv1.VirtualMachine, dv cdiv1.DataV
 	}
 
 	vmSpec.Spec.Template.Spec.Volumes = append(vmSpec.Spec.Template.Spec.Volumes, volume)
-	vmSpec.Spec.Template.Spec.Domain.Devices.Disks = append(vmSpec.Spec.Template.Spec.Domain.Devices.Disks, kubevirtDisk)
+	disks := append(vmSpec.Spec.Template.Spec.Domain.Devices.Disks, kubevirtDisk)
+	sort.Slice(disks, func(i, j int) bool {
```
jakub-dzon, line 306: Why is this sorting needed?

pkg/providers/vmware/client/client.go
```diff
@@ -129,14 +143,11 @@ func (r RichVmwareClient) GetVMHostProperties(vm *object.VirtualMachine) (*mo.Ho
 }
 
 // StartVM requests VM start and doesn't wait for it to complete.
-func (r RichVmwareClient) StartVM(id string) error {
+func (r RichVmwareClient) StartVM(moRef string) error {
```
jakub-dzon, line 144: The tests are still using UUID; please update them.

pkg/providers/vmware/client/client.go
```diff
@@ -151,14 +162,11 @@ func (r RichVmwareClient) StartVM(id string) error {
 }
 
 // StopVM stops the VM and waits for the vm to be stopped.
-func (r RichVmwareClient) StopVM(id string) error {
+func (r RichVmwareClient) StopVM(moRef string) error {
```
jakub-dzon, line 163: The tests are still using UUID; please update them.

pkg/providers/vmware/client/client.go
```diff
@@ -80,6 +88,9 @@ func (r RichVmwareClient) getVMByUUID(id string) (*object.VirtualMachine, error)
 	if err != nil {
 		return nil, err
 	}
+	if vmRef == nil {
```
jakub-dzon, line 92: I understand that both err and vmRef are nil when the vm is not found; please cover it in tests.

pkg/providers/vmware/client/client.go
```diff
@@ -94,6 +105,9 @@ func (r RichVmwareClient) getVMByInventoryPath(vmPath string) (*object.VirtualMa
 	if err != nil {
 		return nil, err
 	}
+	if vm == nil {
```
jakub-dzon: Please cover this case in tests as well.

pkg/providers/vmware/provider.go
```diff
@@ -0,0 +1,422 @@
...
+	vmProperties          *mo.VirtualMachine
+	vmiObjectMeta         metav1.ObjectMeta
+	vmiTypeMeta           metav1.TypeMeta
+	vmwareClient          *vclient.RichVmwareClient
+	vmwareSecretDataMap   map[string]string
+}
+
+// NewVmwareProvider creates a new VmwareProvider
+func NewVmwareProvider(vmiObjectMeta metav1.ObjectMeta, vmiTypeMeta metav1.TypeMeta, client client.Client, tempClient *tempclient.TemplateV1Client, factory pclient.Factory,
+	_ kvConfig.KubeVirtConfigProvider, ctrlConfig ctrlConfig.ControllerConfig) VmwareProvider {
```
jakub-dzon: `_ kvConfig.KubeVirtConfigProvider` is unused; please remove it.

pkg/providers/vmware/provider.go
```diff
@@ -0,0 +1,422 @@
...
+		osFinder:              &osFinder,
+		templateHandler:       templates.NewTemplateHandler(templateProvider),
+		templateFinder:        vtemplates.NewTemplateFinder(templateProvider, osFinder),
+	}
+}
+
+// Init initializes the VmwareProvider with a given credential secret and VirtualMachineImport
+func (r *VmwareProvider) Init(secret *corev1.Secret, instance *v1beta1.VirtualMachineImport) error {
+	r.vmwareSecretDataMap = make(map[string]string)
+	err := yaml.Unmarshal(secret.Data[vmwareSecretKey], &r.vmwareSecretDataMap)
```
jakub-dzon, line 93: How will the rest of the code behave when there are missing required entries in the map? Will the user know what to do to fix the problem? 
Please document expected behaviour with tests.

pkg/providers/vmware/provider.go
```diff
@@ -0,0 +1,422 @@
...
+func (r *VmwareProvider) StopVM(cr *v1beta1.VirtualMachineImport, client client.Client) error {
+	vmwareClient, err := r.getClient()
+	if err != nil {
+		return err
+	}
+	vm, err := r.getVM()
+	if err != nil {
+		return err
+	}
+	err = vmwareClient.StopVM(vm.Reference().Value)
```
jakub-dzon: Provider's `StopVM` method is called in every reconciliation and `vmwareClient.StopVM` always retrieves VM from VMware, which we already have in provider. In single migration one vm import is reconciled many times, and when multiplied by number of VMs in mass migration it can give really big number of requests to VMware API. Is it possible to check the power status here?

pkg/providers/vmware/provider.go
```diff
@@ -0,0 +1,422 @@
...
+		ownerreferences.NewVMImportOwnerReference(r.vmiTypeMeta, r.vmiObjectMeta),
+	}
+	err := r.secretsManager.CreateFor(&newSecret, vmiName)
+	if err != nil {
+		return nil, err
+	}
+	return &newSecret, nil
+}
+
+func foldErrors(errs []error, vmiName k8stypes.NamespacedName) error {
```
jakub-dzon: Seems like a good candidate for moving to utils package; ovirt provider defines exactly the same function.

pkg/providers/vmware/provider.go
```diff
@@ -0,0 +1,422 @@
...
+	vmwareClient, err := r.getClient()
+	if err != nil {
+		return err
+	}
+	return vmwareClient.TestConnection()
+}
+
+// Validate checks whether the source VM and resource mapping is valid.
+func (r *VmwareProvider) Validate() ([]v1beta1.VirtualMachineImportCondition, error) {
+	// TODO: implement vmware rule validation
```
jakub-dzon, line 316: Will both source VM compatibility validation and resource mapping be implemented in the external service?

pkg/providers/vmware/provider.go
```diff
@@ -0,0 +1,422 @@
...
+	return []v1beta1.VirtualMachineImportCondition{
+		conditions.NewCondition(v1beta1.Valid, string(v1beta1.ValidationCompleted), "Validation completed successfully", corev1.ConditionTrue),
```
jakub-dzon, line 318: That method needs to set both `v1beta1.Valid` and `v1beta1.MappingRulesVerified` conditions, where `Valid` is responsible for resource mapping checks and `MappingRulesVerified` for source VM checks.

pkg/providers/vmware/provider_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/vim25/mo"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	kubevirtv1 "kubevirt.io/client-go/api/v1"
+)
+
+var _ = Describe("Processing a template", func() {
```
jakub-dzon, line 465: Please implement tests for other `Provider` interface methods which are not no-ops.

jakub-dzon (changes requested):
It seems to be possible to implement big chunk of negative e2e tests; please add them.

pkg/providers/vmware/mapper/mapper.go
mansam: Since the the import controller is iterating over a map of DVs, MapDisk gets called for each DV in a nondeterministic order which results in the disks being in an arbitrary order. This sort ensure the disks are attached in the same order as the devices on the source VM.

pkg/providers/vmware/provider.go
mansam: `vmwareClient.StopVM` uses `getByMoRef` to get the VM, which does not involve an API call-- it just constructs a VM object from the reference. `vm.PowerState()` does however involve a vSphere API call each time. I'll see about using the vmProperties that the provider has already collected to check the power state while avoiding the API call.

pkg/providers/vmware/provider.go
mansam: This was a placeholder until the validation rules are implemented. I'll remove it until then.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -830,17 +831,22 @@ func isIncomplete(condition *v2vv1.VirtualMachineImportCondition) bool {
...
 	// The type of the provider is evaluated based on the source field from the CR
 	if vmi.Spec.Source.Ovirt != nil {
-		config, err := r.ctrlConfigProvider.GetConfig()
-		if err != nil {
-			log.Error(err, "Cannot get controller config.")
-		}
 		provider := ovirtprovider.NewOvirtProvider(vmi.ObjectMeta, vmi.TypeMeta, r.client, r.ocClient, r.factory, r.kvConfigProvider, config)
 		return &provider, nil
 	}
+	if vmi.Spec.Source.Vmware != nil {
```
masayag, line 851: There is a small chance that both ovirt and vmware might be provided (if it isn't enforced by the schema validation - could be validated by 'oneOf'). In that case we should deny the request.

pkg/providers/vmware/mapper/mapper.go
jakub-dzon: :+1:

pkg/providers/vmware/provider.go
jakub-dzon: :+1:

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
machacekondra: It would be nice to note this in documentation, that only one provider at time is supported.

pkg/providers/vmware/mapper/mapper.go
machacekondra: Can you please explain it in the comment in the code?

machacekondra, line 306: Can you please explain it in the comment in the code?

pkg/providers/vmware/provider.go
mansam: @fdupont-redhat do you know whether the shared validation service will handle mappings?

ghost: @mansam probably not in MVP, but in my opinion it should handle both. The inventory service will collect both VMware and Kubevirt data, so it will be able to request a validation for the mapping.

mansam:
@jakub-dzon these test failures seem unrelated to my code. Do you have any idea what is causing them?

jakub-dzon:
That is a timeout while waiting for a temporary namespace to be removed as part of the test teardown. Did you try running the tests locally? Maybe something is somehow blocking the ns removal after your changes and the problem is reproducible locally.

mansam:
I hadn't previously tried running the oVirt tests locally all the way through since they'd take a very, very long time to complete. I've got them running now, but I expect them to take hours.

I did write a Vmware version of one of the failing oVirt tests, `should fail when ImportWithoutTemplate feature gate is disabled and VM template can't be found`, which failed with the same `Timeout waiting for the condition` error. I checked the controller logs, and it appears that CleanUp failed due to the VM not having been created yet, which caused a reconciler error, which left the test waiting for a condition that never arrived. I made a local change to allow CleanUp to pass without an error if the VM doesn't exist yet, but then started getting permissions errors regarding the service account accessing the templates.

mansam:
The tests are passing locally, so I am at a loss for what is wrong. Unfortunately the automated tests seem to swallow the real issue, showing just the timeout.

jakub-dzon:
I increased the namespace removal timeout just to see if that helps. 
It would be good to see what is blocking the namespace removal (get all the information about it and it's objects when the error is encountered using `KubeVirtClient` or `runKubectlCommand`).

jakub-dzon:
The error is still there for oVirt, but the e2e job is reported as successful (probably because TestVmwareProviderE2E has been run last and didn't report any errors).

mansam:
One of the tests failed locally with the same timeout. The namespace is stuck at 'terminating', with the following status:

```
  - lastTransitionTime: "2020-09-10T01:56:49Z"
    message: All content successfully deleted, may be waiting on finalization
    reason: ContentDeleted
    status: "False"
    type: NamespaceDeletionContentFailure
  - lastTransitionTime: "2020-09-10T01:56:49Z"
    message: 'Some resources are remaining: virtualmachineimports.v2v.kubevirt.io
      has 1 resource instances'
    reason: SomeResourcesRemain
    status: "True"
    type: NamespaceContentRemaining
  - lastTransitionTime: "2020-09-10T01:56:49Z"
    message: 'Some content in the namespace has finalizers remaining: vmimport.v2v.kubevirt.io/restore-state
      in 1 resource instances'
    reason: SomeFinalizersRemain
    status: "True"
    type: NamespaceFinalizersRemaining
  phase: Terminating
```

The `virtualmachineimport` that it seems to be hung on has a value set for deletionTimestamp, and has the finalizer in question:
```
    deletionTimestamp: "2020-09-10T01:56:49Z"
    finalizers:
    - vmimport.v2v.kubevirt.io/restore-state
```

I don't see any finalizer messages in the controller log that pertain to that namespace or its resources however.

jakub-dzon:
The finalizer should be removed by this piece of code:
https://github.com/kubevirt/vm-import-operator/blob/f569b17d50faae625ad3064d20434aaef3ed879c/pkg/providers/ovirt/provider.go#L378
All the errors should be collected and then formatted here:
https://github.com/kubevirt/vm-import-operator/blob/f569b17d50faae625ad3064d20434aaef3ed879c/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go#L1013
Any logs contain  the prefix `Import failure clean-up for:`?

mansam:
I don't see any with that prefix in my controller log.

mansam:
It looks like the last commit allowed the tests to pass for both oVirt and Vmware. (https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_vm-import-operator/388/pull-vm-import-operator-e2e-k8s-1.18/1304034525679980544) I've force pushed one other change, just waiting for the tests to pass again.

mansam:
@jakub-dzon @pkliczewski The tests have finally all passed. Any further changes that need to be made, or can this be merged?

pkliczewski:
It looks good to me. @jakub-dzon it is your call. @mansam what about https://github.com/kubevirt/containerized-data-importer/pull/1320 ? Do we need one more PR?

mansam:
@pkliczewski Yes, at least one more PR will be necessary to add the virt-v2v job and to incorporate the VDDK datasource from kubevirt/containerized-data-importer#1320.

automation/execute-tests.sh
```diff
@@ -23,15 +23,18 @@ echo "Using: "
...
 FAKEOVIRT_CA_PATH=${FAKEOVIRT_CA_PATH:-$(pwd)/_out/fakeovirt-ca.pem}
 
 $KUBECTL -n "$IMAGEIO_NAMESPACE" exec deploy/imageio-deployment -c imageiotest -- cat /tmp/certs/ca.pem > "$FAKEOVIRT_CA_PATH"
 
 $KUBECTL apply -f tests/os-mapping/os-mapping.yaml
 
-go test ./tests/ovirt --v -timeout 120m -kubeconfig "$KUBECONFIG" -ovirt-ca "$FAKEOVIRT_CA_PATH" -imageio-namespace "$IMAGEIO_NAMESPACE" -kubevirt-namespace "$KUBEVIRT_NAMESPACE" -default-sc "$DEFAULT_SC" -nfs-sc "$NFS_SC"
\ No newline at end of file
+go test ./tests/ovirt --v -timeout 120m -provider ovirt -kubeconfig "$KUBECONFIG" -ovirt-ca "$FAKEOVIRT_CA_PATH" -imageio-namespace "$IMAGEIO_NAMESPACE" -kubevirt-namespace "$KUBEVIRT_NAMESPACE" -default-sc "$DEFAULT_SC" -nfs-sc "$NFS_SC"
+go test ./tests/vmware --v -timeout 120m -provider vmware -kubeconfig "$KUBECONFIG" -kubevirt-namespace "$KUBEVIRT_NAMESPACE" -default-sc "$DEFAULT_SC" -nfs-sc "$NFS_SC"
```
jakub-dzon: If the tests for oVirt fail and for VMware don't nobody will notice unless one looks into the prow logs; like here: https://prow.apps.ovirt.org/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_vm-import-operator/388/pull-vm-import-operator-e2e-k8s-1.18/1303949338703040512. Would it behave correctly if we run `go test ./tests --v ...`?

mansam: I've pushed a new commit to make the provider flag unnecessary on the command line, and to run both sets of tests in the same command with `go test ./tests/ovirt ./tests/vmware`. That looks like it should work.

jakub-dzon (approved):
Great work! 

I've fixed a few of comments and run auto formatting - please remember to run `make` before committing; it formats files in a standard way.

PR #390, opened 2020-09-04
Allow specifying nodeSelector, affinity and tolerations

pkliczewski:
As part of placement rules we need to extend vm import config and update controller deployment.


pkg/apis/v2v/v1alpha1/vmimportconfig_types.go
```diff
@@ -25,6 +25,33 @@ import (
 // VMImportConfigSpec defines the desired state of VMImportConfig
 type VMImportConfigSpec struct {
 	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
+
+	// Rules on which nodes controller pod(s) will be scheduled
+	Infra NodePlacement `json:"infra,omitempty"`
```
jakub-dzon, line 31: Should we touch v1alpha any more? We have newer v1beta.

jakub-dzon (commented):
Looks good - just one question.

pkg/apis/v2v/v1alpha1/vmimportconfig_types.go
pkliczewski: This is what I saw in CDI so I did the same. In the past we had to fix `v1alpha` to have successful HCO deployment.

jakub-dzon: OK

jakub-dzon (approved):

jakub-dzon (changes requested):
Please fix the CI build.

pkg/apis/v2v/v1alpha1/vmimportconfig_types.go
masayag, line 31: should this field be mandatory ?

pkg/apis/v2v/v1beta1/vmimportconfig_types.go
```diff
@@ -25,6 +25,33 @@ import (
 // VMImportConfigSpec defines the desired state of VMImportConfig
 type VMImportConfigSpec struct {
 	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
+
+	// Rules on which nodes controller pod(s) will be scheduled
+	Infra NodePlacement `json:"infra,omitempty"`
```
masayag, line 31: should this field be mandatory ?
Also, what 'Infra' stands for in this context ? perhaps just 'Placement' or 'NodePlacement' which describes the nature of this field as is ?

jakub-dzon (approved):

pkg/apis/v2v/v1alpha1/vmimportconfig_types.go
pkliczewski: Based on HCO design my understanding is that if it is not there we rely on scheduler to place the controller pod on any node. In this context it is OK to have it optional.

pkg/apis/v2v/v1beta1/vmimportconfig_types.go
pkliczewski: As @jakub-dzon is working on extracting common code from both CDI and vm-import I used the same name of the type as CDI to make it easy to extract.

pkg/apis/v2v/v1alpha1/vmimportconfig_types.go
pkliczewski: Added  // +optional  for both

pkg/operator/controller/controller.go
```diff
@@ -775,6 +776,9 @@ func (r *ReconcileVMImportConfig) getOperatorArgs(cr *v2vv1.VMImportConfig) *Ope
 		if cr.Spec.ImagePullPolicy != "" {
 			result.PullPolicy = string(cr.Spec.ImagePullPolicy)
 		}
+		if &cr.Spec.Infra != nil {
```
masayag, line 779: If you define Infra 
`NodePlacement `json:"infra,omitempty"``
as
`Infra *NodePlacement `json:"infra,omitempty"``

you could just check 
```
if cr.Spec.Infra != nil {
...
```

without need to check the address with &. This seems to complete the API notation of having an optional element.
Is there any specific reason not to define the NodePlacement on the VMImportConfig as *NodePlacement ?

pkliczewski: Wtih you suggestion compiler complains:
> pkg/operator/controller/controller.go:779:20: cannot convert nil to type "github.com/kubevirt/vm-import-operator/pkg/apis/v2v/v1beta1".NodePlacement

pkliczewski: The only reason is consistency with CDI.

masayag: I'm referring to the guidelines on https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

masayag: > The only reason is consistency with CDI.

okay  - so we can stick to it.

masayag (approved):

Issue #218, opened 2020-05-15
Add a step to convert the VM


ghost:
If the source virtual machine provider is not KVM/VirtIO based, e.g. VMware, we should have a step to convert the newly created virtual machine. This should happen between `ImportDisks` and `StartVM`. This cannot be done during the transfer phase where each disk transfer is independent. During the conversion, all the devices, disks and NICs, must be present to correctly create the internal devices definitions in the operating system. 

From an implementation perspective, this could be a job with a pod that mounts all the PVCs under a specific path. This pod would be responsible for the conversion and will most likely wrap `virt-v2v` command to report the progress in a specific format.

pkliczewski:
The PR is merged, closing

PR #400, opened 2020-09-14
Add convertGuest step to run virt-v2v on vSphere VMs

mansam:
This pull request adds a step called `convertGuest` that follows `importDisks`, which runs a virt-v2v batch `Job` to convert the disks. As part of this process it creates a `ConfigMap` containing a libvirt domain XML file that the virt-v2v job uses to locate the disks.


hack/generate-manifests.sh
```diff
@@ -42,6 +44,7 @@ rendered=$( \
 	--operator-version=${IMAGE_TAG} \
 	--operator-image="${CONTAINER_PREFIX}/${OPERATOR_IMAGE}:${IMAGE_TAG}" \
 	--controller-image="${CONTAINER_PREFIX}/${CONTROLLER_IMAGE}:${IMAGE_TAG}" \
+  --virtv2v-image="${CONTAINER_PREFIX}/${VIRTV2V_IMAGE}:${IMAGE_TAG}" \
```
pkliczewski: please align with other arguments

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -343,39 +399,308 @@ func (r *ReconcileVirtualMachineImport) addWatchForImportPod(instance *v2vv1.Vir
...
+	if err != nil {
+		return false, err
+	}
+
+	configMap, err := r.findLibvirtDomainConfigMap(vmName)
+	if err != nil {
+		return false, err
+	}
+	if configMap == nil {
+		configMap, err = r.makeLibvirtDomainConfigMap(instance, vmSpec)
```
pkliczewski: When do we clean it up? All the resources we create should be deleted once the vm is imported.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -343,39 +399,308 @@ func (r *ReconcileVirtualMachineImport) addWatchForImportPod(instance *v2vv1.Vir
...
+	if job.Status.Active > 0 {
+		return false, nil
+	}
+
+	if job.Status.Failed > 0 {
+		err := r.endGuestConversionAsFailed(provider, instance, "virt-v2v job failed")
+		if err != nil {
+			return false, err
+		}
+		return false, nil
```
pkliczewski, line 434: Should conversion failure fail the import? Here I see that the job was created and failed we do not fail the conversion and the job won't get recreated because of line 427. In this case the import would stuck. Do I understand the code correctly?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -343,39 +399,308 @@ func (r *ReconcileVirtualMachineImport) addWatchForImportPod(instance *v2vv1.Vir
...
+					GenerateName: virtV2vJobName + "-",
+					Namespace:    vmSpec.Namespace,
+				},
+				Spec: v1.PodSpec{
+					RestartPolicy: v1.RestartPolicyNever,
+					Containers: []v1.Container{
+						{
+							Name:            virtV2vJobName,
+							Image:           virtV2vImage,
+							ImagePullPolicy: v1.PullIfNotPresent,
```
pkliczewski: `ImagePullPolicy` should be provided we should not hardcode it. There is a param that we can use here.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -49,25 +52,30 @@ import (
 )
 
 const (
-	sourceVMInitialState = "vmimport.v2v.kubevirt.io/source-vm-initial-state"
+	AnnAPIGroup          = "vmimport.v2v.kubevirt.io"
```
jakub-dzon: Please comment exported constant.

pkliczewski (changes requested):
Please add functional test(s) to cover entire flow.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -290,14 +311,49 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+		}
+
+		if !done {
+			return reconcile.Result{}, nil
+		}
 	}
 
-	if err = r.startVM(provider, instance, vmName); err != nil {
-		return reconcile.Result{}, err
+	if shouldConvertGuest(instance) {
```
jakub-dzon: I think that provider should return proper value instead of checking `instance.Spec.Source.Vmware != nil` in `shouldConvertGuest` (the status condition check belongs to the controller)

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -343,39 +399,308 @@ func (r *ReconcileVirtualMachineImport) addWatchForImportPod(instance *v2vv1.Vir
...
+		}
+	}
+
+	job, err := r.findGuestConversionJob(vmName)
+	if err != nil {
+		return false, err
+	}
+	// the job doesn't exist, so create it
+	if job == nil {
+		job = r.makeGuestConversionJobSpec(vmSpec)
```
jakub-dzon: I think batch job also needs to be removed during clean-up (in case job gets stuck and import fails for some other reason).

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
jakub-dzon: Line 785 should fail the whole import.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -343,39 +399,308 @@ func (r *ReconcileVirtualMachineImport) addWatchForImportPod(instance *v2vv1.Vir
...
+	volumeMounts = append(volumeMounts, v1.VolumeMount{
+		Name:      vmSpec.Name,
+		MountPath: "/mnt/v2v",
+	})
+	return volumes, volumeMounts
+}
+
+// findLibvirtDomainConfigMap finds the libvirt domain xml configmap for a given VM, returning nil if it can't be found.
+func (r *ReconcileVirtualMachineImport) findLibvirtDomainConfigMap(vmName types.NamespacedName) (*v1.ConfigMap, error) {
+	configMapList := &v1.ConfigMapList{}
```
jakub-dzon: What do you think about using https://github.com/kubevirt/vm-import-operator/blob/master/pkg/configmaps/configmaps.go?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -343,39 +399,308 @@ func (r *ReconcileVirtualMachineImport) addWatchForImportPod(instance *v2vv1.Vir
...
+}
+
+// makeLibvirtDomainConfigMap creates a libvirt domain xml configmap for a VM to be used during guest conversion
+func (r *ReconcileVirtualMachineImport) makeLibvirtDomainConfigMap(instance *v2vv1.VirtualMachineImport, vmSpec *kubevirtv1.VirtualMachine) (*v1.ConfigMap, error) {
+	domain := r.makeLibvirtDomain(vmSpec)
+	domxml, err := xml.Marshal(domain)
+	if err != nil {
+		return nil, err
+	}
+	domainXMLConfigMap := &v1.ConfigMap{
```
jakub-dzon: What do you think about using https://github.com/kubevirt/vm-import-operator/blob/master/pkg/configmaps/configmaps.go? Like in https://github.com/kubevirt/vm-import-operator/blob/2cefbcc19ab4c5e98b12a08a18fa978de9cbb282/pkg/providers/ovirt/provider.go#L467

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -343,39 +399,308 @@ func (r *ReconcileVirtualMachineImport) addWatchForImportPod(instance *v2vv1.Vir
...
+			return false, err
+		}
+		return false, nil
+	}
+
+	return job.Status.Succeeded > 0, nil
+}
+
+// findGuestConversionJob finds the guest conversion job for a given VM, returning nil if it can't be found.
+func (r *ReconcileVirtualMachineImport) findGuestConversionJob(vmName types.NamespacedName) (*batchv1.Job, error) {
```
jakub-dzon: Could the whole conversion job creation logic be moved to some other package to take some responsibility away from the controller and make that code more testable (please test it:))?

jakub-dzon (changes requested):

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
mansam: Makes sense. It might also make sense to create a similar package for the Jobs so they follow the same pattern. What do you think?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
mansam: Absolutely. Do you have a preference for where it should live?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
jakub-dzon: Yes, good idea.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
jakub-dzon: Maybe `pkg/jobs`?

pkg/apis/v2v/v1alpha1/virtualmachineimport_types.go
```diff
@@ -194,6 +197,9 @@ const (
 	// CopyingDisks represents the creation of data volumes based on source VM disks
 	CopyingDisks ProcessingConditionReason = "CopyingDisks"
 
+	// ConvertingGuest
```
masayag: represents... (just for consistency with the rest)

pkg/apis/v2v/v1beta1/virtualmachineimport_types.go
```diff
@@ -221,6 +224,9 @@ const (
 	// CopyingDisks represents the creation of data volumes based on source VM disks
 	CopyingDisks ProcessingConditionReason = "CopyingDisks"
 
+	// ConvertingGuest
```
masayag: represents... (just for consistency with the rest)

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -343,39 +399,308 @@ func (r *ReconcileVirtualMachineImport) addWatchForImportPod(instance *v2vv1.Vir
...
+		// when the instance is removed.
+		if err := controllerutil.SetControllerReference(instance, job, r.scheme); err != nil {
+			return false, err
+		}
+		err := r.client.Create(context.TODO(), job)
+		if err != nil {
+			return false, err
+		}
+
+		processingCond := conditions.NewProcessingCondition(string(v2vv1.ConvertingGuest), "Running virt-v2v", corev1.ConditionTrue)
```
masayag: for debug purpose, would it be useful to include the v2v job name as part of the message for this condition "Running virt-v2v job-vm-name..." ?
which will allow the user that monitors the CR to get the job namespaced-name quickly ?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -343,39 +399,308 @@ func (r *ReconcileVirtualMachineImport) addWatchForImportPod(instance *v2vv1.Vir
...
+			return false, err
+		}
+	}
+
+	if job.Status.Active > 0 {
+		return false, nil
+	}
+
+	if job.Status.Failed > 0 {
+		err := r.endGuestConversionAsFailed(provider, instance, "virt-v2v job failed")
```
masayag: same about including the job name in this message.
in case of a job failure - the job resource should remain to allow further debug or the failure reason should be propagated to the user via CR condition and/or via controller log.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -343,39 +399,308 @@ func (r *ReconcileVirtualMachineImport) addWatchForImportPod(instance *v2vv1.Vir
...
+	}
+
+	return job.Status.Succeeded > 0, nil
+}
+
+// findGuestConversionJob finds the guest conversion job for a given VM, returning nil if it can't be found.
+func (r *ReconcileVirtualMachineImport) findGuestConversionJob(vmName types.NamespacedName) (*batchv1.Job, error) {
+	jobList := &batchv1.JobList{}
+	matchingLabels := client.MatchingLabels(map[string]string{VMLabel: vmName.Name})
+	err := r.client.List(context.TODO(), jobList, matchingLabels)
```
masayag: namespace isn't needed in this case?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -725,29 +1058,37 @@ func shouldStartVM(instance *v2vv1.VirtualMachineImport) bool {
 	return instance.Spec.StartVM != nil && *instance.Spec.StartVM && conditions.HasSucceededConditionOfReason(instance.Status.Conditions, v2vv1.VirtualMachineReady)
 }
 
-// manageDataVolumeState update current state according to progress of import
-func (r *ReconcileVirtualMachineImport) manageDataVolumeState(instance *v2vv1.VirtualMachineImport, dvsDone map[string]bool, numberOfDvs int) (bool, error) {
+func shouldConvertGuest(instance *v2vv1.VirtualMachineImport) bool {
```
masayag: I'd prefer to encapsulate this Provider - since this is provider specific and the controller after the point in which it creates the providers doesn't evaluate the Spec.Source for the provider type.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -725,29 +1058,37 @@ func shouldStartVM(instance *v2vv1.VirtualMachineImport) bool {
...
 }
 
 func (r *ReconcileVirtualMachineImport) createDataVolume(provider provider.Provider, mapper provider.Mapper, instance *v2vv1.VirtualMachineImport, dv cdiv1.DataVolume, vmName types.NamespacedName) error {
+	instanceNamespacedName := types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}
+	// Update condition to create VM:
+	processingCond := conditions.NewProcessingCondition(string(v2vv1.CopyingDisks), "Copying virtual machine disks", corev1.ConditionTrue)
+	err := r.upsertStatusConditions(instanceNamespacedName, processingCond)
+	if err != nil {
+		return err
+	}
```
masayag, line 853: That's identical to lines 723-729. can it be shared via a function ?

masayag (changes requested):

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
mansam: Replacing this entirely with a Job manager in the same vein as the other resource managers.

mansam:
I'd like to add more tests around the `convertGuest` step, but I haven't been able to figure out how to test that it's setting appropriate Conditions and such without just testing that the mocks do what I told them to do, since the client is mocked out entirely and that's what's doing most of the work.

pkg/providers/vmware/provider.go
```diff
@@ -389,10 +410,7 @@ func (r *VmwareProvider) prepareDataVolumeCredentials() (*mapper.DataVolumeCrede
 }
 
 func (r *VmwareProvider) ensureSecretIsPresent(keyAccess, keySecret string) (*corev1.Secret, error) {
-	vmiName := k8stypes.NamespacedName{
-		Name:      r.vmiObjectMeta.Name,
-		Namespace: r.vmiObjectMeta.Namespace,
-	}
+	vmiName := r.getNamespacedName()
```
jakub-dzon, line 407: :+1:

pkg/jobs/jobs.go
```diff
@@ -0,0 +1,75 @@
...
+	prefix       = "vmimport.v2v.kubevirt.io"
+	vmiNameLabel = prefix + "/vmi-name"
+)
+
+// Manager provides operations on config maps
+type Manager struct {
+	client client.Client
+}
+
+// NewManager creates new config map manager
```
jakub-dzon: "... new batch job manager"

pkg/jobs/jobs.go
```diff
@@ -0,0 +1,75 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+
+)
+
+const (
+	prefix       = "vmimport.v2v.kubevirt.io"
+	vmiNameLabel = prefix + "/vmi-name"
+)
+
+// Manager provides operations on config maps
```
jakub-dzon: "on batch jobs"

pkg/providers/provider.go
```diff
@@ -34,6 +35,9 @@ type Provider interface {
 	CleanUp(bool, *v2vv1.VirtualMachineImport, rclient.Client) error
 	FindTemplate() (*oapiv1.Template, error)
 	ProcessTemplate(*oapiv1.Template, *string, string) (*kubevirtv1.VirtualMachine, error)
+	NeedsGuestConversion() bool
+	GetGuestConversionJob() (*batchv1.Job, error)
```
jakub-dzon, line 39: Is the guest conversion job tailored only for Vmware? Can it be used as is for other providers? If it is universal, maybe it should be implemented outside of this package and called from the controller?

jakub-dzon (changes requested):

pkg/providers/provider.go
ghost: In the near future, it won't be used for other providers. The next provider will likely be OpenStack and it doesn't require a conversion as it is based on KVM.

jakub-dzon: :+1:

mansam: Since the managers are owned by the provider the creation of the configmap and job resources needs to be done there, but the portion that generates the job spec and the libvirt domain xml can be extracted into a shared location.

jakub-dzon:
@mansam - please remember to run `make` to format the code in a standard way.

jakub-dzon (approved):

mansam:
Rebased to fix go.sum conflict.

pkliczewski:
@mansam Please rebase and make sure CI is green

pkliczewski (approved):

pkliczewski:
@masayag Please take a look when you have some time

mansam:
Retesting because the vddk-test image has been updated.

/retest

masayag (approved):

ghost:
Fixes https://github.com/kubevirt/vm-import-operator/issues/218

PR #404, opened 2020-09-16
Update CDI to 1.23.1 and use VDDK datasource in vSphere mapper

mansam:
The first commit updates CDI to 1.23.1 and generates a client for it at `pkg/generated/cdi`. `framework.go` is updated to use this generated client. 
The second commit uses the updated CDI to replace the placeholder datasource with the proper VDDK datasource in the vSphere mapper.


jakub-dzon (changes requested):
It looks that now it should be possible to implement an e2e test that imports vmware machine (and doesn't start it); please add it.

hack/update-codegen.sh
```diff
@@ -34,3 +34,10 @@ bash "${CODEGEN_PKG}"/generate-groups.sh "client" \
   --fake-clientset=false \
   --output-base "$SCRIPT_ROOT/../../../" \
   --go-header-file "${SCRIPT_ROOT}"/hack/boilerplate.go.txt
+
+bash "${CODEGEN_PKG}"/generate-groups.sh "client" \
```
pkliczewski, line 38: We need to use generic client for kubevirt in the future. It is not the issue with CDI but with kubevirt.

pkliczewski:
@jakub-dzon please rebase

tests/vmware/basic_net_vm_import_test.go
```diff
@@ -0,0 +1,111 @@
...
+		created, err := f.VMImportClient.V2vV1beta1().VirtualMachineImports(namespace).Create(&vmi)
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(created).To(BeSuccessful(f))
+
+		retrieved, _ := f.VMImportClient.V2vV1beta1().VirtualMachineImports(namespace).Get(created.Name, metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		vmBlueprint := v1.VirtualMachine{ObjectMeta: metav1.ObjectMeta{Name: retrieved.Status.TargetVMName, Namespace: namespace}}
+		Expect(vmBlueprint).To(BeRunning(f))
```
jakub-dzon, line 56: Is that expected that the machine gets into running state without conversion? Is it the case here that the VM as a machine starts, but the system does not boot? 

It would be good to have a way of verifying whether the OS is working properly (on a separate PR, maybe in #400 - when it should actually work).

jakub-dzon (approved):
Looks good. Just one question/comment to keep in mind during further development.

pkliczewski (approved):

PR #406, opened 2020-09-18
Validate template memory requirements

pkliczewski:
Bug-Url: https://bugzilla.redhat.com/1871433

### DIFF
--- a/pkg/providers/ovirt/provider.go
+++ b/pkg/providers/ovirt/provider.go
@@ -263,7 +263,7 @@ func (o *OvirtProvider) Validate() ([]v2vv1.VirtualMachineImportCondition, error
 		return []v2vv1.VirtualMachineImportCondition{}, errors.New("VM has not been loaded")
 	}
 	vmiName := o.GetVmiNamespacedName()
-	return o.validator.Validate(vm, &vmiName, o.resourceMapping), nil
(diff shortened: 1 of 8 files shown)


pkg/providers/ovirt/provider.go
```diff
@@ -263,7 +263,7 @@ func (o *OvirtProvider) Validate() ([]v2vv1.VirtualMachineImportCondition, error
 		return []v2vv1.VirtualMachineImportCondition{}, errors.New("VM has not been loaded")
 	}
 	vmiName := o.GetVmiNamespacedName()
-	return o.validator.Validate(vm, &vmiName, o.resourceMapping), nil
+	return o.validator.Validate(vm, &vmiName, o.resourceMapping, o.templateFinder), nil
```
jakub-dzon, line 266: What do you think about not retrieving the template twice and refactoring into something along these lines?:
- retrieve template in the controller before validation - set it on provider;
- in provider pass the template to validation (instead of the template finder) and
  - check whether the template was found (if required) - I think we talked about making it part of validation, not processing;
  - check the memory requirements.

pkliczewski: I have been thinking about it but we handle the template differently in both places. As well as we would always fetched it but in some cases validation would fail so it wouldn't be needed.

jakub-dzon (approved):

machacekondra (approved):

PR #408, opened 2020-09-22
New metric counter for cancelled imports

ydayagi:
Signed-off-by: Yaron Dayagi ydayagi@gmail.com

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -2,6 +2,6 @@
 
 Virtual Machine Import uses Prometheus for metrics reporting. The metrics can be used for real-time monitoring. Virtual Machine Import does not persist its metrics, if a member restarts, the metrics will be reset.
 
-| Name               | Description                                                     | Type    | Labels                       |
-|--------------------|-----------------------------------------------------------------|---------|------------------------------|
-| kubevirt_vmimport  | The total number of successfull/failed Virtual Machine imports. | Counter | result=<successful>/<failed> |
+| Name               | Description                                                               | Type    | Labels                                   |
+|--------------------|---------------------------------------------------------------------------|---------|------------------------------------------|
+| kubevirt_vmimport  | The total number of successfull/failed/cancelled Virtual Machine imports. | Counter | result=<successful>/<failed>/<cancelled> |
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -4,4 +4,4 @@ Virtual Machine Import uses Prometheus for metrics reporting. The metrics can be
 
 | Name               | Description                                                               | Type    | Labels                                   |
 |--------------------|---------------------------------------------------------------------------|---------|------------------------------------------|
-| kubevirt_vmimport  | The total number of successfull/failed/cancelled Virtual Machine imports. | Counter | result=<successful>/<failed>/<cancelled> |
+| kubevirt_vmimport  | The total number of successfull/failed/cancelled Virtual Machine imports. | Counter | result=successful\|failed\|cancelled     |
--- a/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
+++ b/pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
@@ -1128,17 +1128,12 @@ func (r *ReconcileVirtualMachineImport) updateProgress(instance *v2vv1.VirtualMa
 
 func (r *ReconcileVirtualMachineImport) afterSuccess(vmName types.NamespacedName, p provider.Provider, instance *v2vv1.VirtualMachineImport) error {
 
-	err := utils.RemoveFinalizer(instance, utils.CancelledImportFinalizer, r.client)
-	if err != nil {
-		reqLogger := log.WithValues("Request.Namespace", instance.Namespace, "Request.Name", instance.Name)
-		reqLogger.Error(err, "Failed to remove finalizer "+utils.CancelledImportFinalizer)
-
-	}
+	r.removeFinalizer(utils.CancelledImportFinalizer, instance)
 
 	metrics.ImportCounter.IncSuccessful()
 	vmiName := types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}
 	var errs []error
-	err = p.CleanUp(false, instance, r.client)
+	err := p.CleanUp(false, instance, r.client)
 	if err != nil {
 		errs = append(errs, err)
 	}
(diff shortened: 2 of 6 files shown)


pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -232,6 +232,37 @@ func (r *ReconcileVirtualMachineImport) Reconcile(request reconcile.Request) (re
...
+	// Add finalizer to handle cancelled import
+	if !r.vmImportInProgress(instance) {
+		err := utils.AddFinalizer(instance, utils.CancelledImportFinalizer, r.client)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Handle deleted import
+	if instance.DeletionTimestamp != nil {
```
pkliczewski, line 244: We already check for deletion timestamp. Please unify and extract a method.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
```diff
@@ -1096,10 +1127,18 @@ func (r *ReconcileVirtualMachineImport) updateProgress(instance *v2vv1.VirtualMa
 }
 
 func (r *ReconcileVirtualMachineImport) afterSuccess(vmName types.NamespacedName, p provider.Provider, instance *v2vv1.VirtualMachineImport) error {
+
+	err := utils.RemoveFinalizer(instance, utils.CancelledImportFinalizer, r.client)
```
pkliczewski: This code look the same as from line 1159. Please extract to a method.

pkliczewski (changes requested):
Please add tests

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
ydayagi: It is a call to a function and then short error handling. If we create a function for such duplication we'll end up with many functions, no?

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
ydayagi: We check after we have a provider. This is for the only existing finalizer. However, this new finalizer does not rely on anything but the cluster itself. Therefore, I can not unify. I tried. I can do something very artificial and in my opinion not so pretty. I can create a single function that takes all possible arguments. Inside in case provider is nil do part A and if not nil do part B. I think this is not such a good approach.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: I noticed duplication maybe we can do it higher up the call chain to remove the need having it in both places.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: My suggestion was to unify cancellation logic. With your PR we are handling it in 2 places. I think we could have something like:

```go
// CR cancelled
if instance.DeletionTimestamp != nil {
 // handle both
}
```

Whether externalizing it or keeping in reconcile it is up to you. I think writing tests for this logic would help us understand which approach would be better.

ydayagi:
> Please add tests

Do we have tests for failed and successful counters?

pkliczewski:
> Do we have tests for failed and successful counters?

The original [implementation](https://github.com/kubevirt/vm-import-operator/pull/328) did not contain tests. In this case we add finalizes that may have implication on the entire flow. This is why I think we need to make sure they are handled correctly.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
ydayagi: I'll do what you tell me, but I think it's OK to have duplicity. I do not know a way to avoid it completely. In this case, for example, these are operations we perform when import is done. If you look at the code that calls afterSuccess and afterFailure you'll see more duplicity (e.g. calling updateProgress(DONE)).

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
ydayagi: I will add tests, of course.
Regarding the handling of cancellation. It is just in this location.
I add finalizer when resource is first handled.
Remove it when it is done.
Call it if resource was deleted.
The other test for deletion is for fializer that requires provider.

pkg/controller/virtualmachineimport/virtualmachineimport_controller.go
pkliczewski: I am open to other approaches. We could refactor the code and remove duplicity.

ydayagi: I can add a new function: onImportDone({fail|success}).
Update progress.
Remove finalizer
Log
...

pkliczewski (commented):
Overall it looks good. It would be great if you could add one or two more tests for `CancelledImportFinalizer`.

ydayagi:
I added 5 tests:
#### Test 1
	Reconcile
		Delete resource not in progress
			counter cancel incremented
			finalizer does not exist
#### Test 2
	Reconcile
		Delete resource in progress
			counter cancel incremented
			finalizer removed
#### Test 3
	afterSuccess
		counter success incremented
		finalizer removed
#### Test 4
	afterFailed
		counter fail incremented
		finalizer removed
#### Test 5
	Reconcile
		Delete resource that is done
			counter cancel not incremented
			finalizer does not exist

masayag (approved):
LGTM, @machacekondra pls review.

pkg/metrics/metrics.go
```diff
@@ -36,12 +37,36 @@ type importCounter struct {
...
 	ic.importCounterVec.With(prometheus.Labels{"result": "successful"}).Inc()
 }
+
+func (ic *importCounter) GetSuccessful() (float64, error) {
+	return ic.getValue(prometheus.Labels{"result": "successful"})
+}
+
+// IncCancelled increment successfull label
+func (ic *importCounter) IncCancelled() {
+	ic.importCounterVec.With(prometheus.Labels{"result": "cancelled"}).Inc()
```
machacekondra, line 67: Would be nice to document `canceled` here: https://github.com/kubevirt/vm-import-operator/blob/master/docs/metrics.md

machacekondra (commented):
Looks ok to me, but please document the `cancelled` metric.

pkliczewski (approved):
Looks good to me. I think we need to wait on stable branch creation before merging this PR.

pkg/metrics/metrics.go
ydayagi: done

docs/metrics.md
```diff
@@ -2,6 +2,6 @@
 
 Virtual Machine Import uses Prometheus for metrics reporting. The metrics can be used for real-time monitoring. Virtual Machine Import does not persist its metrics, if a member restarts, the metrics will be reset.
 
-| Name               | Description                                                     | Type    | Labels                       |
-|--------------------|-----------------------------------------------------------------|---------|------------------------------|
-| kubevirt_vmimport  | The total number of successfull/failed Virtual Machine imports. | Counter | result=<successful>/<failed> |
+| Name               | Description                                                               | Type    | Labels                                   |
+|--------------------|---------------------------------------------------------------------------|---------|------------------------------------------|
+| kubevirt_vmimport  | The total number of successfull/failed/cancelled Virtual Machine imports. | Counter | result=<successful>/<failed>/<cancelled> |
```
masayag: when [previewed](https://github.com/kubevirt/vm-import-operator/blob/1c6a2747fcc5c91f50e5426cabf43ceb943560ac/docs/metrics.md) , the content of the 'Labels' column is trimmed. Can you fix it ?

ydayagi: It was like that since beginning. Fixed it.

masayag (approved):

machacekondra (approved):

PR #413, opened 2020-10-01
Update placement types

pkliczewski:
Bug-Url: https://bugzilla.redhat.com/1884142

### DIFF
--- a/pkg/operator/resources/operator/operator.go
+++ b/pkg/operator/resources/operator/operator.go
@@ -442,47 +442,48 @@ func CreateVMImportConfig() *extv1.CustomResourceDefinition {
 											},
 										},
 										"infra": {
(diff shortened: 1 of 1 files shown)


jakub-dzon (commented):
There are no functional tests for that feature. Did you confirm it works as expected?
Please add functional test to make sure it does and won't break in the future.

pkliczewski:
> There are no functional tests for that feature. Did you confirm it works as expected?
> Please add functional test to make sure it does and won't break in the future.

Yes, I tested it with HCO and it works. I would like to add test but in different PR. I tried to generate a client for `vmimportconfig` but the generator do not works as expected. I need to debug

jakub-dzon (approved):
:+1:

PR #419, opened 2020-10-13
Do not override machine type for i440fx

pkliczewski:
Custom emulated machine is used to override machine type we should not allow to happen when i440fx is used.

Bug-Url: https://bugzilla.redhat.com/1868073

### DIFF
--- a/pkg/providers/ovirt/mapper/mapper.go
+++ b/pkg/providers/ovirt/mapper/mapper.go
@@ -515,7 +515,10 @@ func (o *OvirtMapper) mapArchitecture() *kubevirtv1.Machine {
 
 	// Override machine type in case custom emulated machine is specified
 	if custom, ok := o.vm.CustomEmulatedMachine(); ok {
-		machine.Type = custom
+		// do not override when i440fx which is unsupported
+		if !strings.Contains(custom, "i440fx") {
+			machine.Type = custom
+		}
 	}
 
 	return machine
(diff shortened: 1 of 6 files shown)


masayag (approved):

pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -515,7 +515,10 @@ func (o *OvirtMapper) mapArchitecture() *kubevirtv1.Machine {
 
 	// Override machine type in case custom emulated machine is specified
 	if custom, ok := o.vm.CustomEmulatedMachine(); ok {
-		machine.Type = custom
+		// do not override when i440fx which is unsupported
+		if !strings.Contains(custom, "i440fx") {
```
jakub-dzon, line 519: Shouldn't it be part of VM validation (blocking or warning)? Will the target VM work properly if we ignore `CustomEmulatedMachine`?

pkliczewski: Interesting question. Machine type is not mandatory so we should not block this in validation. Setting the value base on cpu architecture works ok.

jakub-dzon: I think we need to inform user that we are going to ignore that setting by reporting log or warn validation failure.

pkliczewski: Make sense, let's add validation rule and log that we ignore.

pkliczewski:
No conditions got created on VMI and test failed. It seems like the CI was overloaded.

jakub-dzon (approved):

PR #421, opened 2020-10-15
Use Privileged ServiceAccount for guest conversion

mansam:
This adds a key to the vm-import-controller-config configmap, `privilegedServiceAccount.name`. When creating the guest conversion job the provider will check for the presence of the privileged service account in the target namespace. If it exists, the job
will be run as that SA so that it can mount `/dev/kvm` on the host. This significantly speeds up the guest conversion process.

### DIFF
--- a/pkg/guestconversion/guestconversion.go
+++ b/pkg/guestconversion/guestconversion.go
@@ -20,7 +20,7 @@ var (
 // MakeGuestConversionJobSpec creates a job spec for a virt-v2v job,
 // containing a volume and a mount for each volume on the VM, as well
 // as a volume and mount for the config map containing the libvirt domain XML.
-func MakeGuestConversionJobSpec(vmSpec *v1.VirtualMachine, libvirtConfigMap *corev1.ConfigMap) *batchv1.Job {
+func MakeGuestConversionJobSpec(vmSpec *v1.VirtualMachine, libvirtConfigMap *corev1.ConfigMap, privilegedSA string) *batchv1.Job {
 	// Only ever run the guest conversion job once per VM
 	completions := int32(1)
 	parallelism := int32(1)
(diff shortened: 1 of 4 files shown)


ghost (changes requested):
Would it be possible to add a test where `privilegedSA` is not empty?

ghost:
@mansam, I forgot to give the example to add the hostPath volume. In the `makeJobVolumeMounts` methods, you'll need to add:

```
        if privilegedSA != "" {
                // add volume and mount for /dev/kvm in order to have better performance.
                // the node selector of the pod will ensure that /dev/kvm exists on the node.
                volumes = append(volumes, corev1.Volume{
                        Name: "kvm",
                        VolumeSource: corev1.VolumeSource{
                                HostPath: &corev1.HostPathVolumeSource{
                                        Path: "/dev/kvm",
                                },
                        },
                })
                volumeMounts = append(volumeMounts, corev1.VolumeMount{
                        Name:      "kvm",
                        MountPath: "/dev/kvm",
                })
        }
```

And of course pass `privilegedSA` as an argument.

pkliczewski (commented):
It would be great to have a functional test for this flow.

masayag (dismissed):

pkg/guestconversion/guestconversion.go
```diff
@@ -20,15 +20,15 @@ var (
 // MakeGuestConversionJobSpec creates a job spec for a virt-v2v job,
 // containing a volume and a mount for each volume on the VM, as well
 // as a volume and mount for the config map containing the libvirt domain XML.
-func MakeGuestConversionJobSpec(vmSpec *v1.VirtualMachine, libvirtConfigMap *corev1.ConfigMap) *batchv1.Job {
+func MakeGuestConversionJobSpec(vmSpec *v1.VirtualMachine, libvirtConfigMap *corev1.ConfigMap, privilegedSA string) *batchv1.Job {
```
masayag, line 23: The privilegedSA is a concern of the GuestConversionJob.
Can it be obtained within guestconversion.MakeGuestConversionJobSpec instead of passing it from the provider ?
It will required passing the k8s client to this method.

PR #422, opened 2020-10-19
Expose /dev/kvm in the guest conversion pod

mansam:
Uses the KVM device plugin to expose /dev/kvm to the guest conversion pod.

Replaces https://github.com/kubevirt/vm-import-operator/pull/421

### DIFF
--- a/pkg/guestconversion/guestconversion.go
+++ b/pkg/guestconversion/guestconversion.go
@@ -6,6 +6,7 @@ import (
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	v1 "kubevirt.io/client-go/api/v1"
 	libvirtxml "libvirt.org/libvirt-go-xml"
 )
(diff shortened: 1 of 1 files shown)


pkliczewski:
Couple of vmware tests failed with:

> {type: Succeeded, status: False, reason: DataVolumeCreationFailed, message: Error while importing disk image: . Data volume vm-import-e2e-tests-resource-mapping-wcxf9/target-vm-disk-202-1 creation failed: admission webhook "datavolume-validate.cdi.kubevirt.io" denied the request:  Missing Data volume source}

@mansam please check

pkg/guestconversion/guestconversion.go
```diff
@@ -42,9 +43,22 @@ func MakeGuestConversionJobSpec(vmSpec *v1.VirtualMachine, libvirtConfigMap *cor
 							Image:           virtV2vImage,
 							VolumeMounts:    volumeMounts,
 							ImagePullPolicy: imagePullPolicy,
+							// Request access to /dev/kvm via Kubevirt's Device Manager
+							Resources: corev1.ResourceRequirements{
+								Requests: corev1.ResourceList{
```
ghost: IIUC, the `Requests` is not necessary. In my tests, it `Limits` was enough.

mansam: Indeed, according to https://kubevirt.io/api-reference/master/definitions.html#_k8s_io_api_core_v1_resourcerequirements it sets the Requests to be equal to Limits if Requests is not set. Thus, this is redundant and can be removed.

pkg/guestconversion/guestconversion.go
```diff
@@ -42,9 +43,19 @@ func MakeGuestConversionJobSpec(vmSpec *v1.VirtualMachine, libvirtConfigMap *cor
 							Image:           virtV2vImage,
 							VolumeMounts:    volumeMounts,
 							ImagePullPolicy: imagePullPolicy,
+							// Request access to /dev/kvm via Kubevirt's Device Manager
+							Resources: corev1.ResourceRequirements{
```
masayag, line 47: @fdupont-redhat Is there a possibility for this request to fail and how will the conversion behave in that case ?

ghost: If this request fails, the scheduler will not be able to schedule the pod and the job will sit there until it's possible.
This resource request is also used by Kubevirt to launch virtual machines, so I guess that they explored this.

ghost: And the NodeSelector makes sure that the pod will be on a node where /dev/kvm is present.

masayag (approved):

jakub-dzon (approved):

machacekondra (approved):

pkliczewski (approved):

PR #425, opened 2020-10-21
Install the latest version of CDI on the test cluster

mansam:
Curling from `/releases/latest` on Github is not reliable because point releases of older versions can appear as the latest release. Currently, `1.18.3` is returned when querying Github for the latest version, which breaks our tests. 

This change gets a list of releases from the API, sorts them by tag, and takes the latest.

### DIFF
--- a/cluster/common.sh
+++ b/cluster/common.sh
@@ -9,7 +9,7 @@ function debug {
 }
 
 function install_cdi {
-    export VERSION=$(curl -s https://github.com/kubevirt/containerized-data-importer/releases/latest | grep -o "v[0-9]\.[0-9]*\.[0-9]*")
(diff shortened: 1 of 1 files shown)


cluster/common.sh
```diff
@@ -9,7 +9,7 @@ function debug {
 }
 
 function install_cdi {
-    export VERSION=$(curl -s https://github.com/kubevirt/containerized-data-importer/releases/latest | grep -o "v[0-9]\.[0-9]*\.[0-9]*")
```
jakub-dzon, line 12: The reason for having that release version dynamic was to detect early that there is incompatibility or bug that needs to be fixed on either side. I wouldn't want to loose that capability. 
Something like following one-liner should be fine (I haven't verified whether it works in all situations/input).
```
curl -s https://api.github.com/repos/kubevirt/containerized-data-importer/releases | jq 'sort_by(.tag_name) | .[-1].tag_name' -r
```

jakub-dzon (changes requested):

jakub-dzon (approved):

PR #426, opened 2020-10-21
Use DeletePropagationForeground when deleting jobs

mansam:
This ensures that the owned pods are cleaned up when the job is removed.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1888588

### DIFF
--- a/pkg/jobs/jobs.go
+++ b/pkg/jobs/jobs.go
@@ -3,9 +3,9 @@ package jobs
 import (
 	"context"
 	"fmt"
-
 	"github.com/kubevirt/vm-import-operator/pkg/utils"
 	batchv1 "k8s.io/api/batch/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


pkg/jobs/jobs.go
```diff
@@ -68,7 +68,9 @@ func (m *Manager) DeleteFor(vmiCrName types.NamespacedName) error {
 		return err
 	}
 	if job != nil {
-		return m.client.Delete(context.TODO(), job)
+		foreground := metav1.DeletePropagationForeground
+		opts := &client.DeleteOptions{PropagationPolicy: &foreground}
+		return m.client.Delete(context.TODO(), job, opts)
```
masayag, line 77: Can it be verified as part of a functional test ? Else, an automation test should verify there are no leftovers.

masayag (approved):

pkg/jobs/jobs.go
mansam: Yes, I'm working on functional tests.

pkliczewski (approved):

PR #428, opened 2020-10-22
vSphere: FindTemplate should search for a desktop template if no server templates are found

mansam:
The vSphere provider FindTemplate method should pass a nil workload to the template provider, since it can't get the workload from the vSphere API and setting a default can result in failing to find a valid template.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1890486

### DIFF
--- a/pkg/providers/vmware/templates/template-finder.go
+++ b/pkg/providers/vmware/templates/template-finder.go
@@ -2,19 +2,22 @@ package templates
 
 import (
 	"fmt"
-	"sort"
-
 	"github.com/kubevirt/vm-import-operator/pkg/providers/vmware/os"
 	"github.com/kubevirt/vm-import-operator/pkg/templates"
 	templatev1 "github.com/openshift/api/template/v1"
 	"github.com/vmware/govmomi/vim25/mo"
+	"sort"
 )
 
 const (
 	// TemplateNamespace stores the default namespace for kubevirt templates
-	TemplateNamespace = "openshift"
+	templateNamespace = "openshift"
 	defaultFlavor     = "medium"
-	defaultWorkload   = "server"
+)
+
+var (
+	serverWorkload  = "server"
+	desktopWorkload = "desktop"
 )
 
 // TemplateFinder attempts to find a template based on given parameters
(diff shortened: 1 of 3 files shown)


pkg/providers/vmware/templates/template-finder.go
```diff
@@ -40,8 +39,7 @@ func (f *TemplateFinder) FindTemplate(vm *mo.VirtualMachine) (*templatev1.Templa
 	// We update metadata from the source vm so we default to medium flavor
 	namespace := TemplateNamespace
 	flavor := defaultFlavor
-	workload := defaultWorkload
-	tmpls, err := f.templateProvider.Find(&namespace, &os, &workload, &flavor)
+	tmpls, err := f.templateProvider.Find(&namespace, &os, nil, &flavor)
```
jakub-dzon: With `nil` workload the provider will return multiple matching templates, from which only the newest would be picked. I think that resigning from controlling the workload type will lead to less predictable behaviour and application of suboptimal templates (with CPU flags tuned to a completely different workload) to imported VMs. 
I can imagine that user adds a new "desktop"-like template for a certain OS to the configuration and later all VMware imports for that OS use that template, even if most of them require `server`-type optimizations.

`server` is generally a quite safe default and I know that it has similar (but probably less severe) issues as described above (i.e. performance of Windows desktop VMs might suffer).

Maybe introduction of an optional `workload` property to the VirtualMachineImport would help here? It would give the user control over the workload type, if needed.

@pkliczewski, @machacekondra, @masayag, @fdupont-redhat - what do you think? Maybe it's actually not a real issue.

pkliczewski: I agree with @jakub-dzon. It is important to use proper workload type since optimizations part of the template are based on the workload type. I like the idea with workload property or having default workloads for specific OS types.

pkliczewski (changes requested):
Please consider different approach to finding correct workload.

pkg/providers/vmware/templates/template-finder.go
mansam: The problem with setting `server` as the default is that it fails to find appropriate images for e.g. windows 10, which may only have server templates. I can search for a `server` template first, and fall back to searching for a `desktop` one if that fails. Is that more acceptable?

mansam: @pkliczewski @jakub-dzon  as a stopgap until a workload property can be added or default workloads can be determined for each OS, would it be acceptable to default to looking for a `server` template, and then fall back to looking for a `desktop` template if nothing is returned?

pkliczewski: @mansam I think we could try with this approach. @jakub-dzon what do you think?

masayag: > Maybe introduction of an optional `workload` property to the VirtualMachineImport would help here? It would give the user control over the workload type, if needed.

I'd prefer exposing template identifier (name + namespace) instead of specific properties for identifying the template.
Else, we'd end up with exposing the templates attributes used as field selectors.

Without having a template specified by the user, @mansam's approach to use `server` and to fallback to `desktop` sounds good to me.

jakub-dzon: I'm okay with that approach. Moti's suggestion to pass template identifier instead of workload makes more sense than just the workload; it can provide easy workaround for any potential template resolution issues.

pkliczewski: As far as I remember we were explicitly asked not to use the name. @omeryahud Do you have any suggestions how to handle this?

omeryahud: > As far as I remember we were explicitly asked not to use the name. @omeryahud Do you have any suggestions how to handle this?

Yes, using template names is generally not a good idea, because the names of the latest templates may change as new template bundles are released, and if you allow the user to choose the template used, they might not pick the latest template.

If you decide to accept template properties from the user, it should be the os/workload/flavor labels.

Just FYI, we are currently working on some things that were mentioned here:

- Default template for each OS - https://issues.redhat.com/browse/CNV-6153
- Having fixed template names - https://issues.redhat.com/browse/CNV-6624

mansam:
Not sure why there's an ovirt test failing, none of that code's been touched by this PR.

/retest

jakub-dzon (approved):

pkliczewski (approved):

masayag (approved):

PR #429, opened 2020-10-22
vSphere: always map machine type to q35

mansam:
This overwrites the machine type that might be set by the template the VM is based on, in order to avoid the VM failing to start due to an unsupported machine type.

### DIFF
--- a/pkg/providers/vmware/mapper/mapper.go
+++ b/pkg/providers/vmware/mapper/mapper.go
@@ -41,6 +41,11 @@ const (
 	networkTypePod    = "pod"
 )
 
+// architectures
+const (
+	q35 = "q35"
+)
+
 var biosTypeMapping = map[string]*kubevirtv1.Bootloader{
 	"efi":  {EFI: &kubevirtv1.EFI{}},
 	"bios": {BIOS: &kubevirtv1.BIOS{}},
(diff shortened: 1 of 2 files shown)


pkg/providers/vmware/mapper/mapper.go
```diff
@@ -427,6 +432,7 @@ func (r *VmwareMapper) MapVM(targetVmName *string, vmSpec *kubevirtv1.VirtualMac
 	nameParts := strings.Split(hostname, ".")
 	vmSpec.Spec.Template.Spec.Hostname = nameParts[0]
 
+	vmSpec.Spec.Template.Spec.Domain.Machine = kubevirtv1.Machine{Type: q35}
```
jakub-dzon, line 435: Please have it verified/"documented" by a unit test so that wouldn't get changed accidentally later.

jakub-dzon (changes requested):

pkliczewski:
@mansam Please take a look at test failure where:

> Error converting guests: virt-v2v job vmimport.v2v.kubevirt.iorzll9 failed

we need to understand what is the reason.

mansam:
Still trying to determine the cause of this test failure. The tests pass on my development machine, and all of the test vms use the same disk images via the fake vddk, so if guest conversion passes for one it should pass for all of them.

pkliczewski (approved):

jakub-dzon (approved):

PR #438, opened 2020-11-16
Add optional AccessMode to StorageResourceMappingItem

mansam:
konveyor/virt-controller would like to be able to specify a supported access mode when creating storage/disk mappings. This adds an `AccessMode` field to the `StorageResourceMappingItem`, and adds support for mapping the `AccessMode` and `VolumeMode` in the vSphere provider.

`AccessMode` and `VolumeMode` will default respectively to `ReadWriteOnce` and `Filesystem` in the vSphere provider.
The oVirt provider will take the access mode from the mapping if it's set, otherwise it will determine the access mode as it previously did.


fdupont-redhat:
I don't see how changing this string can impact tests. But CDI is flaky these days. Respawning test just in case.
/retest

pkliczewski (approved):

Issue #216, opened 2020-05-14
Allow warm import in the reconciliation


ghost:
To reduce the downtime of the VM during an import, it would be interesting to have a warm import mechanism. The warm import means that the disks transfer starts while the VM is powered on. Then, when the _cutover_ is requested, the virtual machine is shutdown to quiesce the disks, and the final delta is transferred, then the VM is powered on.

Here is a proposed workflow for warm import:

1. If the warm import field is false, stop the source VM
2. Create the destination VM CR
3. Create the DataVolume CR, with the warm migration set accordingly
4. When cutover is requested (immediately for cold import)
  4.1. If the warm import field is true, stop the source VM
  4.2. Wait for DataVolume to be complete
5. Start the destination VM

The main difference with current implementation are:

- A new boolean field in VirtualMachineImportSpec struct
- A new `warm` field for the `DataVolumeSource*` structs that support warm import (in CDI)
- A change in the `Reconcile` method to check warm import field

PR #446, opened 2020-12-05
Add support for Warm Import

mansam:
Implements support for Warm Import in the main reconcile loop.

The failing operator test is spurious and seems to be due to operator-utils' `GetMissingEntries` not handling pointers to `metav1.Time`. The times in the VirtualMachineImportStatus also generate the same errors, but the test ignores the status subresource so it's not noticed.

See https://github.com/kubevirt/vm-import-operator/issues/216


mansam:
CDI 1.28 is out, rerunning tests.

/retest

mansam:
@fdupont-redhat Prow is passing now. The Travis failure is a false alarm as I mentioned above.

PR #451, opened 2020-12-10
oVirt mapper: use 'scsi' bus for disks with the 'virtio_scsi' type

mansam:
This sets the destination disk bus type to `scsi` when oVirt reports that the source disk is `virtio_scsi`. Kubevirt does not appear to accept `virtio_scsi` as a bus type, but it looks like `scsi` should work.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1903585. I don't have an oVirt setup to test against, and since the issue is Windows booting in repair mode I don't see any way to create a functional test.

### DIFF
--- a/pkg/providers/ovirt/mapper/mapper.go
+++ b/pkg/providers/ovirt/mapper/mapper.go
@@ -46,6 +46,9 @@ var (
 	DefaultVolumeMode = corev1.PersistentVolumeFilesystem
 )
 
+// DiskInterfaceModelMapping defines mapping of disk interface models between oVirt and kubevirt domains
+var DiskInterfaceModelMapping = map[string]string{"sata": "sata", "virtio_scsi": "scsi", "virtio": "virtio"}
+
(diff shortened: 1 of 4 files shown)


pkg/providers/ovirt/mapper/mapper.go
```diff
@@ -46,6 +46,9 @@ var (
 	DefaultVolumeMode = corev1.PersistentVolumeFilesystem
 )
 
+// DiskInterfaceModelMapping defines mapping of disk interface models between oVirt and kubevirt domains
+var DiskInterfaceModelMapping = map[string]string{"sata": "sata", "virtio_scsi": "scsi", "virtio": "virtio"}
```
mansam, line 50: Moved here to avoid an import cycle.

pkliczewski (commented):
It would be great if you could add a test or two.

pkliczewski (approved):

PR #452, opened 2020-12-14
Remove limits from VM spec

ghost:
We have been told that settings the limits can have a bad impact until ballooning is supported.
This pull request removes that for RHV only, as it was not set when importing from VMware.

```release-note
Remove "limits" from the VM spec when importing.
```

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1904051


pkliczewski (approved):

pkliczewski (changes requested):
Please fix the build:

> go vet ./pkg/... ./cmd/...
> github.com/kubevirt/vm-import-operator/pkg/providers/ovirt/mapper
> pkg/providers/ovirt/mapper/mapper.go:649:21: undefined: maxMemoryQuantity
> Makefile:85: recipe for target 'vet' failed

PR #462, opened 2021-01-18
Run `make fmt` to fix accumulated lint

mansam:


pkliczewski:
@mansam The same with this PR. It seems like we need to fix this test:

> Discrepancy between CRD and Struct Missing or incorrect schema validation at [/spec/finalizeDate], expected type [object] in CRD file [vmimport-crd]

pkliczewski:
@mansam Please rebase

pkliczewski (approved):

PR #464, opened 2021-01-20
Mount block devices properly on the virt-v2v container

mansam:
Block devices need to be mounted as VolumeDevices rather
than VolumeMounts, otherwise the pod will get
stuck in ContainerCreating because the volumes can't be
attached.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1917908


pkliczewski:
@mansam please rebase

mansam:
The test that's most consistently failing is an oVirt test which is checking to ensure that the VM is deleted when the import fails. It's attempting to get the VM from the API and expecting to see a NotFound error, but it doesn't. My best guess at the moment is that it's hung up on some finalizer, but I am still trying to debug it. In any case, the failure should not be related to this PR.

/retest

pkliczewski (approved):

PR #466, opened 2021-01-22
Skip the job and directly create a pod for guest conversion

mansam:
This uses the `anyuid` SCC to set the fsGroup for the virt-v2v pod to the same one used by the CDI importer pod. This solves a permissions problem that was preventing the virtio overlay from being committed properly when importing a VM's disk to the `nfs` storage class.

The virt-v2v pod needs to be created by the vm-import-controller in order for it to have access to the `anyuid` SCC. If it is created via a Job, the job-controller is responsible for creating it and it will not have access to the needed SCC. Thus, all job related code has been refactored to manage a pod directly.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1904797


ghost (approved):
LGTM

pkliczewski (approved):

PR #468, opened 2021-01-26
Fail the import if the target VM can't be found

mansam:
This change causes the VM import to fail if the target VM can't be found after the point where it should have been created. This will usually be due to the VM being deleted.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1894897

image


pkliczewski:
@mansam please rebase

pkliczewski (approved):

PR #469, opened 2021-01-28
Add overhead to requested size when creating DVs

mansam:
This attempts to get the storage class filesystem overhead map from the CDIConfig. If it can't be found, it uses the default overhead  defined in CDI. The size of PVCs requested by the mapper is increased by the percentage defined in the map for that PVC's storage class, or by the default if no value was specified for that storage class.

https://bugzilla.redhat.com/show_bug.cgi?id=1864577


ghost (approved):
LGTM

PR #470, opened 2021-01-29
Add validation that VMWare Tools are present

mansam:
vSphere VMs can't be shutdown gracefully unless the VMWare tools are present. This adds a validation to ensure that either the tools are present, or the VM is already powered off.

I've also fixed another validation that was accidentally commented out, and added tests for both.


mansam:
Updated the vcsim image.

/retest

ghost:
Forced merge since we only see random oVirt test failures in code that hasn't been touched.

PR #472, opened 2021-02-15
Add annotation to allow retaining the v2v pod after a successful import

mansam:
Adds an annotation `vmimport.v2v.kubevirt.io/retain-conversion-pod` which if set will cause the virt-v2v conversion pod to be retained even after a successful import. This is to permit troubleshooting in a situation where the virt-v2v conversion superficially appears to have succeeded, but the VM fails to start correctly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.x
+  - 1.14.x
 
 env:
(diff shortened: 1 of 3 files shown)


ghost (approved):

mansam:
Image problems resolved, back to dealing with random ovirt failures.

/retest

PR #479, opened 2021-04-01
[BZ 1945121] Update storage overhead formula

ghost:
It seems that the formula should be `roundup(size/(1-overhead), 512)`. Rounding integers doesn't work with `math.Round`.
I've also changed the overhead values to use real numbers that have been calculated separately.

```release-note
[BZ 1945121] Fix overhead calculation formula to match CDI.
```


mansam (approved):
LGTM

ghost:
It's Good Friday today, so I'll merge my own PR since it's been approved.

PR #504, opened 2021-08-06
Remove dependency on obsolete Kubevirt config map

mansam:
Removes the LiveMigration validation, because KubeVirt LiveMigration being enabled or not has no bearing on whether or not an oVirt VM can be migrated to it, so this should not block import.

Makes ImportWithoutTemplate configurable via the vm-import-controller-config config map.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1977277


ghost (approved):

mansam:
These test failures aren't related to the changes in this PR. Merging.

PR #521, opened 2023-03-29
Update importWithoutTemplate flag in the doc

sharafcmr:
Update the doc to use the correct approach for enabling  importWithoutTemplate in kubernetes clusters with newer versions of kubevirt(kubevirt-config configmap is obsolete) 

https://github.com/kubevirt/vm-import-operator/blob/master/docs/design.md
Document https://github.com/kubevirt/vm-import-operator/pull/504

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -211,6 +211,19 @@ metadata:
   namespace: kubevirt
 ```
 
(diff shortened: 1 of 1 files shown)


sudharsanomprakash (approved):
Tested and works fine.

